PR #3, opened 2017-04-18
Alpha v2.0

vitalych:


s2e_env/command.py
```diff
@@ -251,17 +251,55 @@ def handle_common_args(self, options):
         self._env_dir = options['env']
         options.pop('env', ())
 
+        try:
+            with open(self.s2eenv_path()):
+                pass
+        except IOError:
+            raise CommandError('This does not look like an S2E project directory.')
```
adrianherrera: I use the term "project" to refer to analysis targets, and an "environment" to refer to the combination of projects, source code, build files, etc. (see the readme). Maybe this is not the clearest, most optimal approach, but for consistency I think this should say 'This does not look like an S2E environment'.

s2e_env/command.py
```diff
@@ -251,17 +251,55 @@ def handle_common_args(self, options):
...
+    def s2eenv_path(self):
+        return self.env_path('.s2eenv')
+
     def env_path(self, *p):
         """
         Create a path relative to the environment directory.
         """
         return os.path.join(self._env_dir, *p)
 
+    def s2e_install_path(self, *p):
```
adrianherrera: I think `s2e_install_path` and `s2e_project_path` should be renamed to `install_path` and `project_path` for consistency (you already have `source_path`, `image_path`, etc.)

s2e_env/command.py
```diff
@@ -251,17 +251,55 @@ def handle_common_args(self, options):
...
+        """
+        Create a path relative to the S2E install directory
+        """
+        return os.path.join(self._env_dir, 'install', *p)
```
adrianherrera: Instead of `os.path.join` you can just make use of the `env_path` method.

E.g. `return self.env_path('install', *p)`

s2e_env/command.py
```diff
@@ -251,17 +251,55 @@ def handle_common_args(self, options):
...
+
     def env_path(self, *p):
         """
         Create a path relative to the environment directory.
         """
         return os.path.join(self._env_dir, *p)
 
+    def s2e_install_path(self, *p):
+        """
+        Create a path relative to the S2E install directory
```
adrianherrera: Full stops at the end of """ comments :)

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+                            help='The name of the image to build. If empty,'
+                                 ' shows available images', nargs='?')
+        parser.add_argument('-g', '--headless', action='store_true',
                             help='Build the image in headless mode (i.e. '
                                  'without a GUI)')
-        parser.add_argument('-s', '--snapshot', required=False,
-                            default='ready', help='Snapshot name. Defaults to '
-                                                  '\'ready\'')
         parser.add_argument('-m', '--memory', required=False, default=256,
                             type=_ram_type,
```
adrianherrera: This is pointless. You can just do `type=int` and delete `_ram_type`.

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
                             help='Amount of RAM allocated to the image. '
                                  'Defaults to 256 MB')
-        parser.add_argument('-c', '--num-cores', required=False, default=4,
+        parser.add_argument('-c', '--num-cores', required=False, default=2,
                             type=_cpu_cores_type,
```
adrianherrera: `type=int` and delete `_cpu_cores_type`

s2e_env/commands/image_build.py
```diff
@@ -22,249 +20,83 @@
...
-
-    This function then returns a tuple containing:
-        1. The client connect (so that we can communicate further with the VM
-        2. The kernel version string
-    """
-    # Start listening and notify the QEMU thread that it can start the VM
-    sock.listen(1)
-    ev.set()
+def _print_group_error(group_name):
+    print "You must belong to %s in order to build images." % group_name
```
adrianherrera: When you want to use the inbuilt `print`, please `from __future__ import print_function` as the first import and use `print()` as a function, not as a statement. The print statement is deprecated and should not be used

s2e_env/commands/image_build.py
```diff
@@ -22,249 +20,83 @@
...
+    print "Please run the following command, then logout and login:"
```
adrianherrera: For consistency can we please use single quotes ' for strings. I will build this into the pylint config

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
         image_name = options['name']
+        memory = options['memory']
         num_cores = options['num_cores']
+        headless = options['headless']
+        clean = options['clean']
+        archive = options['archive']
+        download = options['download']
 
-        # Only override the system settings if specified
-        if options['headless']:
```
adrianherrera: Please keep this. Other wise the default is to use a GUI, which is not a good assumption and will crash. In the constructor I originally had:

```
def __init__(self):
    super(Command, self).__init__()

    # If we are running without an X session, run QEMU in headless mode
    self._headless = os.environ.get('DISPLAY') is None
```

That way the default mode is based on the environment variable.

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+        rule_name = image_name
 
-        Packer is the tool used to build S2E images. It is downloaded as a zip
-        file which is then unzipped into the project's bin directory.
-        """
-        self.info('Fetching Packer')
+        if archive:
+            rule_name = 'archive'
+            if image_name != 'all':
+                rule_name = os.path.join(self.image_path(), image_name + '.tar.xz')
```
adrianherrera: Do not concatinate strings. Prefer `rule_name = os.path.join(self.image_path(), '%s.tar.xz' % image_name)`

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
-        os.chmod(packer_path, st.st_mode | stat.S_IEXEC)
-
-    def _packer_build(self, template_dir, template_name, snapshot='ready',
-                      num_cores=4):
-        """
-        Runs ``packer build`` on the given template file to create the image.
-        """
-        # Get the packer executable
-        packer = self._get_packer()
+        if not headless:
```
adrianherrera: When you add the constructor back this should become

```
if not self._headless:
```

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+            env['GRAPHICS'] = ''
 
-        # Change into the template directory
-        orig_dir = os.getcwd()
-        os.chdir(template_dir)
+        if clean:
+            args = ['make', '-f', os.path.join(img_build_dir, 'Makefile'), 'clean']
+            try:
+                subprocess.check_call(args, env=env, cwd=self.image_path())
```
adrianherrera: I try to avoid the subprocess module and use the sh module. For consistency can we use sh unless there is a really good reason to use subprocess?

```python
try:
    make = sh.Command('make').bake(file=os.path.join(img_build_dir,
                                                     'Makefile'),
                                   directory=self.image_path(),
                                   _out=sys.stdout, _err=sys.stderr,
                                   _env=env_vars, _fg=True)
    if options['clean']:
        make('clean')

    make_image = make.bake(j=num_cores)
    make_image(rule_name)
except ErrorReturnCode as e:
    raise CommandError(e)
```

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
-        finally:
-            # Change back to the original directory
-            os.chdir(orig_dir)
-
-    def _packer_cleanup(self, template_dir):
-        """
-        Cleanup previous Packer builds.
-        """
-        self.info('Cleaning up previous Packer builds')
+        args = [
```
adrianherrera: This can all be replaced by the above code that uses sh :)

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
-        arch = qemu_exe_name.split('-')[-1]
-        libs2e_path = self.env_path('share', 'libs2e', 'libs2e-%s.so' % arch)
-        if not os.path.isfile(libs2e_path):
-            raise CommandError('No libs2e at %s' % libs2e_path)
+        print(' * all - Build all images')
 
-        # Create the libs2e environment variable for taking the snapshot
-        env_vars = os.environ.copy()
-        env_vars['LD_PRELOAD'] = libs2e_path
+        print('\nRun ``s2e image_build <name>`` to build an image. '
```
adrianherrera: `self.info`?

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
-                       ``_run_qemu``).
-            msg: An optional message to display to the user.
+    def _check_core_num(self, value):
+        """
+        Ensure that the number of CPU cores is sensible.
         """
-        if msg:
-            self.info(msg)
+        if value <= 0 or value > 10:
+            self.warn('The specified number of cores seems high')
```
adrianherrera: Can we use a similar warning message to `_check_ram_size`? What is "high" here? Maybe something like:

```python
self.warn('The specified number of cores looks too high. Less than 10 is recommended for best performance.')

File: s2e_env/commands/image_build.py [hunk 16]
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+        images = []
+        if image_name == 'all':
+            images = templates.keys()
+        else:
+            images.append(image_name)
 
-        qemu_thread.daemon = True
-        control_thread.daemon = True
+        for image in images:
+            dest_file = self.image_path(image + '.tar.xz')
```
adrianherrera: `%s.tar.xz` % image

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+            self._download(templates[image]['url'], dest_file)
+            self._decompress(dest_file)
 
-        # Start the threads
-        qemu_thread.start()
-        control_thread.start()
+    def _download(self, url, path):
+        self.info('Downloading ' + url)
```
adrianherrera:
```python
'Downloading %s' % url
```

s2e_env/commands/image_build.py
```diff
@@ -278,413 +110,153 @@ def __init__(self):
...
+        s2e_env.utils.google.download(url, path)
 
-        # Wait for the threads to complete
-        qemu_thread.join()
-        control_thread.join()
+    def _decompress(self, path):
+        self.info('Decompressing ' + path)
+        try:
+            cwd = os.path.dirname(path)
+            subprocess.check_call(['tar', 'xJvf', path], cwd=cwd)
```
adrianherrera: There is no need to use a subprocess here. The following code will work:

```python
import tarfile

def _decompress(self, path):
        self.info('Decompressing %s' % path)
        with tarfile.open(path) as tar:
            tar.extractall(path=os.path.dirname(path))
```

s2e_env/commands/info.py
```diff
@@ -52,28 +50,21 @@ class Command(EnvCommand):
     def handle(self, **options):
         # Naively check if S2E has been built (by checking some QEMU binaries
         # exist)
-        s2e_built = len(glob.glob(self.env_path('bin', 'qemu-system-*'))) > 0
+        s2e_built = len(glob.glob(self.s2e_install_path('bin', 'qemu-system-*'))) > 0
```
adrianherrera: 80 char line limit in Python :)

s2e_env/commands/init.py
```diff
@@ -50,22 +48,38 @@ def add_arguments(self, parser):
         parser.add_argument('dir', help='The environment directory')
         parser.add_argument('-s', '--skip-dependencies', action='store_true',
                             help='Skip the dependency install via apt')
+        parser.add_argument('-b', '--use-existing-install', required=False, default=None,
+                            help='Do not fetch sources but instead use existing S2E installation '
+                                 'whose prefix is specified by this parameter')
```
adrianherrera: do you think "prefix" is the right word here? Does "path" make more sense?

s2e_env/commands/init.py
```diff
@@ -50,22 +48,38 @@ def add_arguments(self, parser):
...
 
+    def install_binary_dist(self, env_path, prefix):
+        self.info('Using S2E installation in %s' % prefix)
+        install = os.path.join(env_path, 'install')
+        shutil.rmtree(install, True)
+        os.symlink(prefix, install)
+
+        # We still need to guest-images repo, because it contains
+        # info about location of the images
+        guest_images_repo = CONSTANTS['repos']['images']['build']
```
adrianherrera: As discussed, we will also need to fetch the s2e-linux-kernel repo at some point (whether here or in image_build)

adrianherrera: Actually can we just move lines 67 - 68 to `_get_img_sources`?

s2e_env/commands/init.py
```diff
@@ -50,22 +48,38 @@ def add_arguments(self, parser):
...
+        prefix = options['use_existing_install']
 
         # First check if we are already in the environment directory
         if os.path.realpath(env_path) == os.path.realpath(os.getcwd()):
             raise CommandError('You cannot create an S2E environment in the '
                                'current working directory. Please `cd ..`')
 
         # Then check if something already exists at the environment directory
-        if os.path.isdir(env_path):
+        if os.path.isdir(env_path) and not os.listdir(env_path) == []:
```
adrianherrera: can just do:

```python
if os.path.isdir(env_path) and os.listdir(env_path):
```

Apparently this is more "pythonic" :)

s2e_env/commands/init.py
```diff
@@ -78,25 +92,34 @@ def handle(self, **options):
...
-        # Get the source repositories
-        self._get_s2e_sources(env_path)
-        self._get_img_sources(env_path)
+        # Create marker file to recognize s2e environment folder
+        with open(os.path.join(env_path, '.s2eenv'), 'w'):
+            pass
+
+        if prefix is not None:
+            self.install_binary_dist(env_path, prefix)
+            return ('Environment created in %s.' % env_path)
```
adrianherrera: No need to () here

s2e_env/commands/init.py
```diff
@@ -182,34 +205,38 @@ def _get_s2e_sources(self, env_path):
...
+
+    def _git_clone_to_source(self, env_path, git_repo):
+        git_url = CONSTANTS['repos']['url']
+
+        git_repo_dir = os.path.join(env_path, 'source', git_repo)
+        git_repo_url = '%s/%s' % (git_url, git_repo)
+        self._git_clone(git_repo_url, git_repo_dir)
+        self.success('Fetched %s' % git_repo)
+
     def _get_img_sources(self, env_path):
```
adrianherrera: Do you think we should fetch s2e-linux-kernel here too?

s2e_env/commands/projects/base.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # Initialize the jinja2 template environment
+        self._template_env = \
+            Environment(loader=FileSystemLoader(TEMPLATES_DIR),
+                        autoescape=False)
+        self._template_env.filters['datetimefilter'] = datetimefilter
+
+    def _validate_binary(self, target_arch, os_name, os_arch):
+        if target_arch == 'x86_64' and os_arch != 'x86_64':
+            raise CommandError('Binary is x86_64 while VM image is %s. Please choose another image.' % os_arch)
```
adrianherrera: 80 chars!

adrianherrera: Also there is no `self` usage here, so this can probably be moved out of the class and just as a top level function

adrianherrera: `os_name` is never used. So just delete it.

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+            '-serial', 'file:serial.txt',
+            '-loadvm', project_desc['image']['snapshot'],
+            '-nographic',
+            '-monitor', 'null',
+            '-m', project_desc['image']['memory'],
+        ] + shlex.split(project_desc['image']['qemu_extra_flags']) + project_args
+
+        return args, env
+
+    def handle(self, **options):
```
adrianherrera: Can we please move this up so that the `add_arguments` and `handle` methods always follow each other. Then other methods come after `handle`

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+        super(Command, self).add_arguments(parser)
+
+        parser.add_argument('project',
+                            help='Project name to run')
+
+        parser.add_argument('project_args', nargs=argparse.REMAINDER,
+                            help='Optional arguments to the S2E launcher script')
+
+        parser.add_argument('-c', '--cores', required=False, default=1,
+                            type=_cpu_cores_type,
```
adrianherrera: just `type=int` and delete `_cpu_cores_type`

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def sigterm_handler(signum=None, frame=None):
+    terminal.print_warn('Got signal %s, terminating S2E' % signum)
+    terminate_s2e()
+
+
+class S2EThread(Thread):
+    def __init__(self, args, env, cwd, stdout, stderr):
+        Thread.__init__(self)
```
adrianherrera: Use `super`. Parent classes should be referenced using:

```python
super(S2EThread, self).__init__()
```

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+        returncode = s2e_main_process.returncode
+        while True:
+            if not _has_s2e_processes(s2e_main_process.pid):
+                break
+            time.sleep(1)
+
+        self._stderr.close()
+        self._stdout.close()
+
+        global terminating
```
adrianherrera: Can you please keep the globals together. I.e.

```python
global s2e_main_process, terminating
```

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+        return True
+    except OSError:
+        return False
+
+
+def terminate_s2e():
+    global terminating
+    terminating = True
+
+    global s2e_main_process
```
adrianherrera: We do not need the `global` statement if we are only reading from the variable. This line can be deleted

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+    # First, send SIGTERM to S2E process group
+    if s2e_main_process is None:
```
adrianherrera: Prefer

```python
if not s2e_main_process:
```

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+    try:
+        os.killpg(s2e_main_process.pid, signal.SIGTERM)
+    except OSError:
+        return
+
+    s2e_main_process.poll()
+
+    # Give S2E time to quit
+    terminal.print_warn('Waiting for S2E processes to quit...')
+    for n in xrange(15):
```
adrianherrera: `n` is never used.


```python
for _ in xrange(15):
```

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+        self._args = args
+        self._env = env
+        self._cwd = cwd
+        self._stdout = stdout
+        self._stderr = stderr
+
+    def run(self):
+        # Launch s2e
+        global s2e_main_process
+        s2e_main_process = subprocess.Popen(
```
adrianherrera: I suspect that we will have to use subprocess here, because you are dealing with signals

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+
+        parser.add_argument('-c', '--cores', required=False, default=1,
+                            type=_cpu_cores_type,
+                            help='Number of cores to run S2E on')
+
+        parser.add_argument('-n', '--no-tui', required=False, default=False,
+                            action='store_true',
+                            help='Disable text UI')
+
+    def _get_project(self, name):
```
adrianherrera: So there is already a `ProjectCommand` class in `s2e_env/command.py` that already has this functionality. You should extend `ProjectCommand` rather than `EnvCommand` here and get this "for free"!

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+
+        self._stderr.close()
+        self._stdout.close()
+
+        global terminating
+        terminating = True
+        return returncode
+
+
+class Command(EnvCommand):
```
adrianherrera: Extend `ProjectCommand` rather than `EnvCommand`

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+        'covered_bbs',
+        'num_crashes',
+        'pov1',
+        'pov2'
+    ]
+
+    def add_arguments(self, parser):
+        super(Command, self).add_arguments(parser)
+
+        parser.add_argument('project',
```
adrianherrera: If you extend `ProjectCommand` you can delete this argument

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+    def _get_project(self, name):
+        path = self.s2e_project_path(name, 'project.json')
+        try:
+            with open(path) as fp:
+                return json.loads(fp.read())
+        except:
+            raise CommandError('Project does not exist or is invalid: ' + path)
+
+    def _setup_env(self, project_desc, project_args, cores, qmp_socket):
+        qemu = self.s2e_install_path('bin', 'qemu-system-' + project_desc['arch'])
```
adrianherrera:
```python
'qemu-system-%s' % project_desc['arch']
```

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def _get_project(self, name):
+        path = self.s2e_project_path(name, 'project.json')
+        try:
+            with open(path) as fp:
+                return json.loads(fp.read())
+        except:
+            raise CommandError('Project does not exist or is invalid: ' + path)
+
+    def _setup_env(self, project_desc, project_args, cores, qmp_socket):
```
adrianherrera: if you extend `ProjectCommand` you do not need to pass `project_desc` as an argument. You can use `self._project_desc` instead

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+    def handle(self, **options):
+        project = options['project']
+        project_args = options['project_args']
+        cores = options['cores']
+        no_tui = options['no_tui']
+
+        # TODO: automatic allocation
+        qmp_socket = ('127.0.0.1', 2014)
+
+        project_desc = self._get_project(project)
```
adrianherrera: not needed with `ProjectCommand` :)

s2e_env/commands/run.py
```diff
@@ -0,0 +1,306 @@
...
+                    time.sleep(1)
+
+            terminate_s2e()
+        finally:
+            qmp_server.shutdown()
+            qmp_server.server_close()
+
+    def _get_data(self):
+        elapsed_time = datetime.datetime.now() - self._start_time
+        binaries = ", ".join(CGCInterfacePlugin.coverage._tb_coverage.keys())
```
adrianherrera: prefer ' to "

adrianherrera: if `_tb_coverage` is being accessed from outside the object, then it should not be marked as "private" (with the underscore). This should be renamed to `tb_coverage`. Pylint will pick this up

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+"""CGCInterface plugin: interacts with S2E"""
+import os
+import logging
+import sys
+
+import coverage
+from coverage import Coverage
+from stats import CGCStats
+
+logger = logging.getLogger(__name__)
```
adrianherrera: This reminds me, I wanted to switch over to the logging module in the commands themselves. Will make it more consistent with your CGC code too!

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+
+sys.path.append(os.path.join(os.path.realpath(__file__), "../"))
```
adrianherrera: "../" should just be '..'

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+            pov_type = data.get('pov_type', None)
+            if pov_type == 1:
+                CGCInterfacePlugin.pov1_count += 1
+            elif pov_type == 2:
+                CGCInterfacePlugin.pov2_count += 1
+        elif testcase_type == 'crash':
+            CGCInterfacePlugin.crash_count += 1
+
+        s2e_coverage_file = data.get('coverage_filename', None)
+        if s2e_coverage_file is not None:
```
adrianherrera:
```python
if not s2e_coverage_file:
```

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @staticmethod
+    def handle_stats(analysis, data):
+        CGCInterfacePlugin.stats.queue_stats(analysis, data)
+
+    @staticmethod
+    def process(data, analysis):
+        data_type = data.get('type', None)
+
+        if data_type not in ['db_error', 'testcase', 'recipe', 'stats']:
```
adrianherrera: Prefer a tuple to a list here

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+            return
+
+        if data_type == 'db_error':
```
adrianherrera: Can we combine this with the previous `if` statement?

s2e_env/commands/server/CGCInterface.py
```diff
@@ -0,0 +1,121 @@
...
+            return
+
+        translate_paths(analysis, data)
+
+        if data_type == 'stats':
+            CGCInterfacePlugin.handle_stats(analysis, data)
+            return
+
+        logger.info('Processing request ' + str(data))
```
adrianherrera: I think the recommended way to use the logging module is

```
logger.info('Processing request %s', data)
```
Note the comma rather than %

s2e_env/commands/server/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+from .CGCInterface import CGCInterfacePlugin
+
+logger = logging.getLogger(__name__)
+
+PLUGINS = {
+    'CGCInterface': CGCInterfacePlugin()
+}
+
+
+class ThreadingMixIn2:
```
adrianherrera: All classes should extend object

s2e_env/commands/server/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+            self.finish_request(request, client_address)
+            self.shutdown_request(request)
+        except:
+            self.handle_error(request, client_address)
+            self.shutdown_request(request)
+
+    def process_request(self, request, client_address):
+        """Start a new thread to process the request."""
+        thread_name = self.thread_name + str(len(self.threads))
+        logger.info('Starting thread "' + thread_name + '"')
```
adrianherrera:
```python
logger.info('Starting thread \'%s\'', thread_name)
```

s2e_env/commands/server/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+                pass
+
+
+class QMPTCPServer(ThreadingMixIn2, SocketServer.TCPServer):
+    allow_reuse_address = True
+    timeout = 5
+    queue = None
+
+    def __init__(self, server_address, RequestHandlerClass):
+        SocketServer.TCPServer.__init__(self, server_address, RequestHandlerClass)
```
adrianherrera: prefer `super`

s2e_env/commands/server/coverage.py
```diff
@@ -0,0 +1,170 @@
...
+        with open(coverage_file) as fp:
+            data = json.load(fp)
+
+        os.remove(coverage_file)
+
+        diff = self.compute_bb_diff(data, coverage_type == TB_COVERAGE)
+
+        logger.info('Processing coverage data diff: %s', str(diff))
+        for module, coverage in diff.iteritems():
+            logger.info('module: %s coverage: %s ', module, str(coverage))
```
adrianherrera: Commas and single quotes! :D

s2e_env/commands/server/coverage.py
```diff
@@ -0,0 +1,170 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+
+import json
+import os
+
+from queueprocessor import QueueProcessor
+import s2e_env.commands
```
adrianherrera: I think something like `from s2e_env.commands import run as run_command` would be better. Then `while not s2e_env.commands.run.terminating:` becomes `while not run_command.terminating:`

s2e_env/commands/server/stats.py
```diff
@@ -0,0 +1,93 @@
...
+
+        while not s2e_env.commands.run.terminating:
+
+            try:
+                # Need timeout to avoid getting stuck on termination
+                item = self._queue.get(True, 2)
+            except:
+                continue
+
+            (analysis, data) = item
```
adrianherrera: No needs for brackets here

s2e_env/commands/server/stats.py
```diff
@@ -0,0 +1,93 @@
...
+            self._global_stats[s] = self._global_stats.get(s, 0) + gs.get(s, 0)
+
+    def run(self):
+        logger.info('Starting stats collection thread')
+
+        while not s2e_env.commands.run.terminating:
+
+            try:
+                # Need timeout to avoid getting stuck on termination
+                item = self._queue.get(True, 2)
```
adrianherrera:
```python
analysis, data = self._queue.get(True, 2)
```

s2e_env/commands/server/stats.py
adrianherrera: Line can be deleted with previous comment

s2e_env/commands/tui/tui.py
```diff
@@ -0,0 +1,274 @@
...
+import time
+import s2e_env.commands
+
+import logging
+logger = logging.getLogger(__name__)
+
+_s_screen = None
+
+
+class Form:
```
adrianherrera: extend object

s2e_env/commands/tui/tui.py
```diff
@@ -0,0 +1,274 @@
...
+    def do_draw(self, ax, ay):
+        self._wnd.mvwin(ay, ax)
+        self._wnd.resize(self._h, self._w)
+        self._wnd.border()
+        self._wnd.refresh()
+
+
+class Label(Form):
+    def __init__(self, parent, x, y, text):
+        Form.__init__(self, parent, x, y, len(text)+2, 1)
```
adrianherrera: prefer super

s2e_env/templates/bootstrap.common.sh
```diff
@@ -0,0 +1,24 @@
+update_guest_tools
```
adrianherrera: Is it okay that the "update_guest_tools" function is not defined in this file?

s2e_env/templates/launch-s2e.sh
```diff
@@ -7,14 +7,54 @@
...
+export S2E_SHARED_DIR={{ install_dir }}/share/libs2e
 export S2E_MAX_PROCESSES=1
+export S2E_UNBUFFERED_STREAM=1
+
+if [ "x$DEBUG" != "x" ]; then
+
+QEMU="$BUILD_DIR/qemu-$BUILD/{{ arch }}-softmmu/qemu-system-{{ arch }}"
+LIBS2E="$BUILD_DIR/libs2e-$BUILD/{{ arch }}-s2e-softmmu/libs2e.so"
+
+cat >> gdb.ini <<EOF
```
adrianherrera: Can we just have `gdb.ini` in its own file in `s2e_env/templates`? Then that file gets copied into the project dir

s2e_env/templates/launch-s2e.sh
```diff
@@ -7,14 +7,54 @@
 # arguments can be passed to this script at run time
 #
 
+INSTALL_DIR="{{ install_dir }}"
+BUILD_DIR="{{ build_dir }}"
+BUILD=debug
+
+if [ "x$1" = "xdebug" ]; then
```
adrianherrera: Do you think it makes sense to separate this into two separate launch scripts? One for debug and one for normal execution?

s2e_env/utils/google.py
```diff
@@ -0,0 +1,77 @@
...
+import terminal
+
+CHUNK_SIZE = 32768
+
+# Inspired from:
+# http://stackoverflow.com/questions/25010369/wget-curl-large-file-from-google-drive
+
+
+def _get_confirm_token(response):
+    for key, value in response.cookies.items():
```
adrianherrera: `value` is never used. Do

```python
for key in response.cookies.keys():
```

s2e_env/utils/google.py
```diff
@@ -0,0 +1,77 @@
...
+                next_count += len(chunk)
+                f.write(chunk)
+
+            if next_count > 1024 * 1024 * 10:
+                terminal.print_info('Downloaded %d bytes' % bytes_count)
+                next_count = 0
+
+
+def _download(docid, destination):
+    url = "https://docs.google.com/uc?export=download"
```
adrianherrera: Do you think this should go in the config file?

s2e_env/templates/launch-s2e.sh
vitalych: I thought about it, then realized that I actually prefer tweaking the launch scripts myself. If you put gdb.ini inline, you can easily change env vars in one place.

s2e_env/templates/bootstrap.common.sh
vitalych: Yeah, it's weird, will fix

s2e_env/templates/launch-s2e.sh
vitalych: It would when it starts getting bigger. We'd have to factor out environment vars setup into a 3rd file to have it common between configs.

adrianherrera:
Ok, sorry, I probably went a bit crazy there. Some high level comments:

* the print function should always be preferred to the print statement (i.e. without brackets). The print statement has been deprecated
* Strings can be either single or double quotes, it's just a matter of preference. However we should pick one and stick with it
* Classes should always inherit from object
* Never do string concatenation

On the structure:

* I think we should move tui out of the commands module. It's not really a command
* can we rename the `s2e_env/commands/projects` directory to `s2e_env/commands/project_creation`? I dunno, that makes more sense to me
* can we move server out of commands. Maybe we should rename it too, not sure
* module names should never use camel case. So can we please rename CGCInterface to cgc_interface

s2e_env/command.py
vitalych: ok

s2e_env/command.py
vitalych: ok

s2e_env/command.py
vitalych: ok. Is this detectable by pylint?

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: I think we should not use print anywhere but logging instead. prints are not nice with the tui stuff.

adrianherrera: Yeah agreed, I mentioned in another comment that I would like to replace the prints with logging

adrianherrera:
Oh can we also update the readme? Maybe we can add a screenshot of the tui, that would be cool!

s2e_env/command.py
adrianherrera: Probably not. I think it's more an Adrian thing

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: good catch, the code is still there, but I forgot the _ so this has no effect anymore.

s2e_env/commands/image_build.py
vitalych: Ok. I agree with the reasons why we shouldn't concatenate, but sometimes it's just much shorter to do so.

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: The point of using subprocess is that you can easily control the process group and terminate S2E. I don't know about sh module, but when calling the shell, you get an extra process in between s2e and the python process, which make things much harder.

vitalych: Oups, sorry, this comment would apply for the s2e run command. It's invalid here as we don't care.

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: ok

s2e_env/commands/image_build.py
vitalych: thanks, didn't know about this.

vitalych: I hope it works with xz

s2e_env/commands/info.py
vitalych: I thought we wanted 120.
80 is questionable, we are long past the old 80 col CRT monitors.

s2e_env/commands/init.py
vitalych: I think prefix is more clear, as it's commonly used in configure scripts to set the installation path.
I can also add an example (e.g., /opt/s2e).

s2e_env/commands/init.py
vitalych: I will do it on-demand in the image build script.

s2e_env/commands/init.py
vitalych: ok

s2e_env/commands/projects/base.py
vitalych: it is used by cgc project. What's the proper way in python to have virtual methods?

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: yes

s2e_env/commands/run.py
vitalych: Good point. I will move the get_project stuff to ProjectCommand.

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/run.py
vitalych: ok

s2e_env/commands/server/CGCInterface.py
vitalych: ok

s2e_env/commands/server/__init__.py
vitalych: ok

s2e_env/commands/server/__init__.py
vitalych: ok

s2e_env/commands/server/__init__.py
vitalych: not sure how to call super here with multiple inheritance

s2e_env/commands/server/coverage.py
vitalych: ok

s2e_env/commands/server/stats.py
vitalych: ok

s2e_env/commands/server/stats.py
vitalych: ok

s2e_env/commands/tui/tui.py
vitalych: ok

s2e_env/commands/tui/tui.py
vitalych: ok

s2e_env/commands/server/__init__.py
adrianherrera: It should still work: http://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance

s2e_env/templates/bootstrap.common.sh
vitalych: ok

s2e_env/utils/google.py
vitalych: it is used

s2e_env/utils/google.py
vitalych: Not sure, this google downloader is supposed to be a black box without any configurable options.

s2e_env/utils/google.py
adrianherrera: Ooh right, so it is! I am blind!

s2e_env/commands/new_project.py
```diff
@@ -631,7 +69,7 @@ def add_arguments(self, parser):
                                  'name of the target program. If you are '
                                  'creating an empty project then this field '
                                  'must be specified')
-        parser.add_argument('-i', '--image', required=True,
+        parser.add_argument('-i', '--image', required=False, default=None,
                             help='The name of an image in the ``images`` '
```
adrianherrera: Could you also please update the help message to explain what will happen if an image is not provided

vitalych: ok

PR #6, opened 2017-04-26
Issue/s2e 200 websvc

vitalych:


s2e_env/commands/run.py
```diff
@@ -202,10 +203,8 @@ def handle(self, *args, **options):
 
         self._start_time = datetime.datetime.now()
 
-        if 'cgc' not in self._project_desc['image']['os_name']:
-            self.warn('Real time statistics only supported for CGC binaries for now.')
-            self.warn('The UI will be empty.')
-            time.sleep(3)
+        if 'cgc' in self._project_desc['image']['os_name']:
+            self._cgc = True
```
adrianherrera:
```python
self._cgc = 'cgc' in self._project_desc['image']['os_name']
```

s2e_env/server/coverage.py
```diff
@@ -174,6 +174,3 @@ def get_tb_coverage(self):
 
     def get_summary(self):
```
adrianherrera: Can you please replace these getters with `@property` dectorators (see https://docs.python.org/2/library/functions.html#property)

E.g. `get_summary` becomes

```python
@property
def summary(self):
    return self._summary
```

Usage then becomes

```python
CollectorThreads.coverage.summary
```

s2e_env/commands/run.py
```diff
@@ -283,6 +282,7 @@ def _get_data(self):
             binaries = "Waiting for analysis to start..."
 
         gs = CollectorThreads.stats.get_global_stats()
```
adrianherrera: Replace with properties

adrianherrera:
Looks good! Just a few minor style things to fix

s2e_env/templates/s2e-config.common.lua
```diff
@@ -0,0 +1,98 @@
+function table.contains(table, element)
```
adrianherrera: I still don't like mixing code with configuration like this. Is it really necessary here?

vitalych: The thing is that we may have different features that require overlapping sets of plugins. I found it more readable to add plugins together with the feature definition. Unfortunately, that means that some plugins may be added twice to the table. So I had to create a function that checks first before adding.

PR #7, opened 2017-04-30
p7zip-full dependency

loverics:
Added 'p7zip-full' as a dependency to install /usr/bin/7z as 'p7zip' only installs /usr/bin/7zr and /usr/bin/p7zip. (Ubuntu 16.04 LTS Server x86-64)

### DIFF
--- a/s2e_env/dat/config.yaml
+++ b/s2e_env/dat/config.yaml
@@ -39,6 +39,7 @@ dependencies:
         - xz-utils
         - docker.io
         - p7zip
+        - p7zip-full
 
         # S2E dependencies
         - libdwarf-dev
--- a/s2e_env/dat/config.yaml
+++ b/s2e_env/dat/config.yaml
@@ -38,7 +38,6 @@ dependencies:
         - python-pip
         - xz-utils
         - docker.io
-        - p7zip
         - p7zip-full
 
         # S2E dependencies


adrianherrera:
Hi Shawn,

Thanks for the PR! If `p7zip-full` is now a dependency, do we still need `p7zip`? Perhaps this can be removed from the config file.

loverics:
Hey Adrian,

I don't think we do. I waiting for s2e build to finish to verify.

Using s2e_env to install and build environments is SO much nicer than
before!

Kind regards,
Shawn

loverics:
Actually I meant s2e init. When creating the docker images the kernel
failed to build without 7z.

adrianherrera:
How did it go?

I did some reading, and it seems that `p7zip` provides the `7zr` command, which is essentially a "light" version of the full `7z` command that is provided by `p7zip-full`. We only use the `7z` command when building an image, so I think that it is safe to remove `p7zip` from the list of dependencies and just leave `p7zip-full`.

Could you please make this change to the PR?

loverics:
Yup no problem. Instead of getting a coffee like the instructions state I
went out for brunch. :) I likely won't get back to this until later this
afternoon.

loverics:
Adrian,

We should be good to go now.

Kind regards,
Shawn

PR #8, opened 2017-04-30
Issue/s2e 211 crc

vitalych:


s2e_env/templates/s2e-config.cgc.lua
```diff
@@ -18,56 +27,103 @@ pluginsConfig.DecreeMonitor = {
     symbolicReadLimitCount = 512,
 }
 
+-------------------------------------------------------------------------------
+-- This plugin exports CGC-specific events to a QMP-based web services.
```
adrianherrera: "to a QMP-based web service"

s2e_env/templates/s2e-config.cgc.lua
```diff
@@ -18,56 +27,103 @@ pluginsConfig.DecreeMonitor = {
...
+-- Note: absence of CFG information is not critical for vulnerability finding.
+
+add_plugin("ControlFlowGraph")
+pluginsConfig.ControlFlowGraph = {
+    reloadConfig = false,
+}
+
+
+-------------------------------------------------------------------------------
+-- This plugin tracks basic block coverage for each state. It required the
```
adrianherrera: "It requires"

s2e_env/templates/s2e-config.lua
```diff
@@ -34,30 +54,84 @@ pluginsConfig.HostFiles = {
...
     },
 }
 
---------------------------------------------------------
+-------------------------------------------------------------------------------
+-- This plugin collects various execution statistics and sends it to a QMP
+-- server that listens on an address:port configured by the S2E_QMP_SERVER
+-- environment variable.
+--
+-- The "s2e run {{ target }}" sets up such a server in order to display stats
```
adrianherrera: The "s2e run {{ target }}" command sets up...

s2e_env/templates/s2e-config.lua
```diff
@@ -66,52 +140,131 @@ pluginsConfig.ModuleExecutionDetector = {
...
         "pc",
     },
     logLevel="info"
 }
 
 {% if use_seeds == true %}
+-------------------------------------------------------------------------------
 add_plugin("MultiSearcher")
+
+-- The SeedSearcher plugin looks for new seeds in the seed directory and
```
adrianherrera: Is this now called SeedScheduler?

s2e_env/commands/run.py
```diff
@@ -246,6 +251,21 @@ def handle(self, *args, **options):
                 qmp_server.shutdown()
                 qmp_server.server_close()
 
+    def _log_to_file(self):
+        # TODO: to put the console log into S2E last, we'll have to create s2e-out*
+        # folders ourselves.
+        log_file = self.project_path('log.txt')
+
+        fileh = logging.FileHandler(log_file, 'w')
+        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
```
adrianherrera: Is there any reason this is a different format to that specified in https://github.com/S2E/s2e-env/blob/master/s2e_env/utils/log.py ? Can we refactor this somehow so that the logging code is all in log.py?

s2e_env/templates/s2e-config.lua
vitalych: No, SeedSearcher still exists and implements the mechanics of fetching seeds. By itself, SeedSearcher doesn't do anything. SeedScheduler implements the policies, i.e., when to schedule seeds. I factored out the policy from the CGCInterface plugin, so that we can also use seeds on Linux.

s2e_env/commands/run.py
vitalych: No, Just copy/pasted random one. I can move this function into log.py.

adrianherrera:
Cool, I've left some comments.

Btw, `sed -i 's/s2e\.py/s2e-env/g'` :)

s2e_env/templates/s2e-config.lua
adrianherrera: Ahh ok cool. That will teach me to briefly look at another PR and make wild assumptions :)

vitalych:
Can't we just call the whole thing s2e.py instead of s2e-env? It sounds cuter.
Reminds me a bit of the team at cgc who called its system crspy (crispy) because their script was called crs.py

s2e_env/utils/log.py
```diff
@@ -35,6 +35,17 @@ def success(self, msg, *args, **kwargs):
         self._log(SUCCESS, msg, args, **kwargs)
 
 
+def log_to_file(log_file):
+    fileh = logging.FileHandler(log_file, 'w')
+    fileh.setFormatter(ColoredFormatter())
```
adrianherrera: If you're logging to a file, it might be best to disable coloured output. The `ColoredFormatter` constructor should allow you to do this

s2e_env/utils/log.py
```diff
@@ -35,6 +35,17 @@ def success(self, msg, *args, **kwargs):
...
+
+    # Reset the root logger in order to replace the default console one
+    log = logging.getLogger()
+    for hdlr in log.handlers[:]:
```
adrianherrera: Is `[:]` necessary? Won't this still work with:

```python
for hdlr in log.handlers:
```

s2e_env/utils/log.py
vitalych: I thought about it, but I like viewing my logs with colors and less -r.

vitalych: I am not against disabling it though, or using a config file setting.

adrianherrera: Nah, let's leave it. I didn't realise you did it on purpose (and also didn't know less had a `-r` flag). We can make it configurable later.

PR #9, opened 2017-05-01
Introduce per environment config files

adrianherrera:
Each S2E environment has its own config file, `s2e.yaml`. Here the user can configure the logging level, logging colours, IDA Pro, etc. This config file replaces the empty `.s2eenv` file that was used to "mark" an S2E environment.


s2e_env/commands/init.py
```diff
@@ -65,13 +65,16 @@ def _install_binary_dist(env_path, prefix):
...
 
-def _install_dependencies():
+def _install_dependencies(install_ida_dependencies=True):
     """
     Install S2E's dependencies.
 
     Only apt-get is supported for now.
     """
-    logger.info('Installing S2E dependencies')
+    if install_ida_dependencies:
```
vitalych: Is this really required in practice? I don't see why anyone would care if there were a couple extra packages installed even if they don't use idapro.

adrianherrera: yeah, you're right. I umm-d and aaah-d over this and decided to play it safe and have it there anyway. But I've removed it now

s2e_env/dat/VERSION
```diff
@@ -1 +1 @@
-0.10
+0.20
```
vitalych: why not use a simple integer? v1,2,3,4, etc.? I don't see much point for x.x notation.

s2e_env/dat/config.yaml
```diff
@@ -19,6 +19,12 @@
 # SOFTWARE.
 
 
+# This is the global configuration file for s2e-env. It describes how all
+# environments are created. Normally a user would not be required to modify
+# the settings in this file. Instead, a separate configuration file is
+# generated for each environment. Users should modify this per-environment
```
vitalych: s/this/the. It's not clear if "this" refers to the global or the local cfg file (especially when taking the sentence out of context, which would happen when quickly skimming through the comment).

setup.py
```diff
@@ -10,7 +33,7 @@
     author='Adrian Herrera',
     author_email='adrian.herrera@epfl.ch',
     version=open(os.path.join('s2e_env', 'dat', 'VERSION'), 'r').read().strip(),
-    url='http://s2e.epfl.ch',
+    url='http://s2e.systems',
```
vitalych: commit msg should read "setup: updated url"

adrianherrera: Fixed

s2e_env/dat/config.yaml
adrianherrera: Fixed

s2e_env/dat/VERSION
adrianherrera: I don't think it should be called v1 until we say it's "officially" out. Isn't it still in alpha?

PR #16, opened 2017-05-07
Issue/s2e 61 windows

vitalych:


s2e_env/commands/image_build.py
```diff
@@ -113,16 +114,48 @@ def _check_vmlinux():
                            'sudo chmod ugo+r /boot/vmlinu*')
 
 
+def _validate_version(descriptor, filename):
+    version = descriptor.get('version', None)
```
adrianherrera: `None` is not required here.

```python
version = descriptor.get('version')
```

s2e_env/commands/image_build.py
```diff
@@ -113,16 +114,48 @@ def _check_vmlinux():
...
+    required_version = CONSTANTS['required_versions']['guest-images']
```
adrianherrera: Please use underscores to keep the config file consistent. I.e. `guest_images`

adrianherrera:
The second commit is very large. There is a lot of refactoring (which is great!) that is unrelated to Windows support. If any of us had done this we would have been told to split it.

adrianherrera:
Can we please merge https://github.com/S2E/s2e-env/pull/15 first. It is much smaller and will be easier to merge before this PR.

s2e_env/commands/image_build.py
```diff
@@ -227,61 +261,89 @@ def handle(self, *args, **options):
...
+            archive_rules = []
+            for r in rule_names:
+                archive_rules.append(os.path.join(self.image_path(), '%s.tar.xz' % r))
+
+            rule_names = archive_rules
+            logger.info('The following archives will be built:')
+            for a in archive_rules:
+                logger.info(' * %s', a)
+
+        self._check_product_keys(templates, image_names)
```
adrianherrera: Do these two methods need to be called for all image build operations?

s2e_env/commands/image_build.py
```diff
@@ -293,6 +355,51 @@ def _clone_kernel(self):
         kernels_repo = CONSTANTS['repos']['images']['linux']
         repos.git_clone_to_source(self.env_path(), kernels_repo)
 
+    def _translate_image_name(self, templates, image_name):
```
adrianherrera: It's not really clear to me what we are "translating". Could you please add a high-level comment for this method.

s2e_env/commands/image_build.py
```diff
@@ -227,61 +261,89 @@ def handle(self, *args, **options):
...
+        for image in image_names:
+            logger.info(' * %s', image)
 
         if options['download']:
-            if image_name == 'all':
-                self.download_images()
-            else:
-                self.download_images(image_name)
-            return 'Successfully downloaded image %s' % image_name
+            self.download_images(image_names)
```
adrianherrera: I'm a little confused, I cannot see this `download_images` method. Is there any reason why it's not `_download_images`?

adrianherrera: Wait never mind, it's in this Mixin class. I don't really like this. Composition is a much, much better solution to this problem, rather than multiple inheritance.

s2e_env/commands/new_project.py
```diff
@@ -42,8 +41,8 @@
 CGC_REGEX = re.compile(r'^CGC 32-bit')
 ELF32_REGEX = re.compile(r'^ELF 32-bit')
 ELF64_REGEX = re.compile(r'^ELF 64-bit')
-PE32_REGEX = re.compile(r'^PE executable')
-PE64_REGEX = re.compile(r'^PE\+ executable')
+PE32_REGEX = re.compile(r'^PE32 executable')
```
adrianherrera: haha cool. I spotted the same mistake :)

s2e_env/commands/project_creation/__init__.py
```diff
@@ -25,227 +26,287 @@
...
+            1. A flag that is ``True`` if a '@@' marker was found and
+               replaced with a symbolic file, or ``False`` if no such
+               marker was found
+            2. A list of the new command-line arguments to render to
+               bootstrap.{bat, sh}
+    """
+    use_symb_input_file = False
+    parsed_args = []
+
+    input_file_substitution = '${SYMB_FILE}'
```
adrianherrera: This is only used once. Just append the string directly to the `parsed_args` list

s2e_env/commands/project_creation/__init__.py
```diff
@@ -25,227 +26,287 @@
...
+        self._project_dir = None
         self._img_json = None
-        self._use_seeds = None
 
-    def _validate_binary(self, target_arch, os_name, os_arch, os_binary_formats):
-        if target_arch == 'x86_64' and os_arch != 'x86_64':
-            raise CommandError('Binary is x86_64 while VM image is %s. Please '
-                               'choose another image.' % os_arch)
+    def handle(self, *args, **options):
+        self._validate_and_create_project(options)
```
adrianherrera: Can we split this method? I.e. have a separate `_validate` method, followed by a `_create_project` method.

s2e_env/dat/config.yaml
```diff
@@ -100,6 +103,10 @@ repos:
         linux: s2e-linux-kernel
         build: guest-images
 
+required_versions:
+    # images.json must have this version
+    guest-images: 2
```
adrianherrera: `guest_images`

adrianherrera: In fact, is this whole `required_versions` thing even necessary?

s2e_env/templates/instructions.txt
```diff
@@ -0,0 +1,82 @@
+Here are some hints to get started with your project:
```
adrianherrera: Good idea moving this into a template. I like it :)

s2e_env/commands/image_build.py
```diff
@@ -217,23 +241,17 @@ def add_arguments(self, parser):
                             help='Deletes all images and rebuild them from '
                                  'scratch')
         parser.add_argument('-a', '--archive', action='store_true',
-                            help='Creates an archive of every supported image')
+                            help='Creates an archive for the specified image')
         parser.add_argument('-d', '--download', action='store_true',
                             help='Download image from the repository instead '
                                  'of building it')
+        parser.add_argument('-i', '--iso-dir',
```
adrianherrera: Do you think instead of specifying the ISO dir that must contain an ISO file with a specific name, we could specify just the path to the ISO itself. So `--iso-dir` becomes `--iso`?

s2e_env/templates/s2e-config.windows.lua
```diff
@@ -0,0 +1,31 @@
...
+-- The dump contains the entire physical memory
+
+add_plugin("BlueScreenInterceptor")
+add_plugin("WindowsCrashDumpGenerator")
+pluginsConfig.WindowsCrashDumpGenerator = {
+    generateCrashDump = true
+}
+
+-------------------------------------------------------------------------------
+-- This plugin collects various bugs from kernel and user space
```
adrianherrera: Is it worth mentioning where the output of this plugin goes? Does it go into the crash dump? Or just written to the log? Or into an additional log file that the user needs to look at?

s2e_env/templates/s2e-config.windows.lua
```diff
@@ -0,0 +1,31 @@
...
+add_plugin("GuestCodePatching")
+pluginsConfig.GuestCodePatching = {
+  moduleNames = {"ntoskrnl.exe", "ntkrnlpa.exe"},
+  allowSelfCalls = true
+}
+
+pluginsConfig.ModuleExecutionDetector['trackAllModules'] = true
+
+-------------------------------------------------------------------------------
+-- This plugin monitors kernel crashes and generates WinDbg crash crash dumps.
```
adrianherrera: "generates WinDbg crash dumps"

adrianherrera: libvmi has a tool to parse these crash dumps doesn't it? It might be worth pointing the user to this tool.

s2e_env/templates/s2e-config.windows.lua
```diff
@@ -0,0 +1,31 @@
...
+-- Guest code patching monitors execution and transparently changes
+-- the target program counter when it encounters a call instructions.
+
+add_plugin("GuestCodePatching")
+pluginsConfig.GuestCodePatching = {
+  moduleNames = {"ntoskrnl.exe", "ntkrnlpa.exe"},
+  allowSelfCalls = true
+}
+
+pluginsConfig.ModuleExecutionDetector['trackAllModules'] = true
```
adrianherrera: Any reason why this config doesn't follow the same "style" as the others.

I.e.

```lua
pluginsConfig.ModuleExecutionDetector= {
    trackAllModules = true
}
```

s2e_env/templates/bootstrap.bat
```diff
@@ -0,0 +1,35 @@
+rem ===== INITIALIZING ENVIRONMENT =====
```
adrianherrera: May as well delete this

adrianherrera:
I cannot comment directly on the line in the file, because it doesn't appear to be a part of this PR, but in `s2e_env/templates/s2e-config.lua` on line 233 it says:

> --   1. SeedSearcher must be used in conjunction with a suitable bootstrap.bat.

I guess this should be bootstrap.sh

s2e_env/templates/s2e-config.windows.lua
adrianherrera: I have some updates to libvmi for these tools (added a README, installs the tools to the bin dir, improves the command-line parsing to make it clearer what the tools do, etc.), but due to CLA things I cannot do anything with them.

I'll hold onto the changes, and when the CLA gets sorted we can revisit them again. Remind me!

adrianherrera:
Some changes in `project.json` have broken the code coverage commands. I get the following error when I run `s2e coverage basic_block ...`

```
File "/home/aherrera/s2e-env/s2e_env/commands/code_coverage/basic_block.py", line 123, in _get_ida_path
 project_arch = self._project_desc['arch'] 
KeyError: 'arch'  
```

Looking at `project.json`, I believe this should be updated to `target_arch`.

adrianherrera:
Also in `basic_block.py` on line 93:

```python
target_path = self._project_desc['target']
```

Should now be

```python
target_path = self._project_desc['target_path']
```

adrianherrera:
In `lcov.py`, line 130:

```python
target_path = self._project_desc['target']
```

is now

```python
target_path = self._project_desc['target_path']
```

adrianherrera:
Since Windows binaries are built automatically, could you please add `g++-mingw-w64-i686`to the package requirements in `config.yaml`. Thanks!

adrianherrera:
Can we fail/warn if a Windows project is created with a filename > 12 chars? This took me slightly over an hour to track down!

s2e_env/commands/image_build.py
vitalych: It's simpler to call them for all, the methods will actually check if the verification is required.

s2e_env/dat/config.yaml
vitalych: It's better to have versions to keep in sync the guest-images and s2e-env repo, it will avoid random crashes due to inconsistencies in the *.json files.

s2e_env/commands/image_build.py
vitalych: It's easier to just have iso dir when you want to build many images at once.

s2e_env/templates/s2e-config.windows.lua
vitalych: the libvmi tool is mostly for debugging, I'd expect people to use WinDbg.

s2e_env/templates/s2e-config.windows.lua
vitalych: Because it adds an extra option to an already existing configuration. Your solution would overwrite the previous config.

s2e_env/commands/project_creation/__init__.py
vitalych: It's a bit tricky. Validate in principle should not modify state, but in this case, it may have to download an image before validating the binary against it. The function is also short enough to not warrant splitting.

s2e_env/commands/image_build.py
vitalych: Indeed, I will add an issue to fix the mixin

s2e_env/templates/s2e-config.windows.lua
adrianherrera: Ahhh ok. I guess I am a Lua newb

adrianherrera:
The code looks fine. I'm just going to create a new env and try to break things.

adrianherrera:
works fine for me!

PR #17, opened 2017-05-10
ExecutionTracer.dat parser module

adrianherrera:
Added a module for parsing execution trace files.

To demonstrate this functionality I also added an `execution_trace` command that takes a trace file and converts it into human-readable JSON. However I don't think this is really the important part (it's just to prove it all works). I think the important part is the parser code.

Let me know what you think!


s2e_env/commands/execution_trace.py
```diff
@@ -0,0 +1,99 @@
...
+class Command(ProjectCommand):
+    """
+    Parses an execution trace into JSON.
+    """
+
+    help = 'Parse an S2E execution trace into JSON.'
+
+    def handle(self, *args, **options):
+        # Get the ExecutionTrace.dat file from s2e-last
+        execution_trace_path = self.project_path('s2e-last', 'ExecutionTracer.dat')
```
vitalych: multicore support?

adrianherrera: ahh ok, good point! I didn't think of that.

vitalych:
Good start! Perhaps add support for multi-core traces and reconstruct the embedded execution tree

adrianherrera:
hey @vitalych, I added multicore support (although haven't tested it yet). Once I aggregate the traces from each node, I sort them by timestamp. Is this what you mean by "reconstruct the embedded execution tree"? I'm not sure what you mean by this.

vitalych:
You need to sort them by their ID (the folder name in which they are contained), then reconstruct the tree by looking at all the fork points. The idea is to have an API that would allow you to get all trace items for path n. You can have a look at the C++ tools for examples.

adrianherrera:
Ok great, thanks! I'll get back to it in the near future

adrianherrera:
I'm going to close this for now. I've been working on it on-and-off every now and then, but it still needs a bit of work. Rather than having it clog the PR backlog, I'll open a new one when it's ready to go.

PR #20, opened 2017-05-18
basic_block_coverage: improvements

adrianherrera:
Occasionally IDA will fail to generate any basic blocks (e.g. due to heavy obfuscation), leading to a divide-by-zero in the statistics calculation. I've added some checks to avoid this.


adrianherrera:
Also updated the JSON format produced by the basic block coverage command. This affects https://github.com/S2E/tools/pull/2 and https://github.com/S2E/docs/pull/9

PR #22, opened 2017-05-22
Issue/s2e 233 refactor mixins

adrianherrera:
* Minor refactoring after running pylint
* Added some comments
* Refactored out mixin classes
* Fixed bug when guessing the guest image


s2e_env/commands/image_build.py
```diff
@@ -46,17 +46,27 @@
...
     groups = [g.gr_name for g in grp.getgrall() if user_name in g.gr_mem]
     gid = pwd.getpwnam(user_name).pw_gid
     groups.append(grp.getgrgid(gid).gr_name)
+
     return groups
 
 
 def _get_user_name():
+    """
+    Get the current user.
```
vitalych: Is this comment necessary? Could Just name function get_current_user().

PR #24, opened 2017-05-24
Added radare support

adrianherrera:
Makes Radare an optional disassembler backend if IDA is unavailable. I will write a tutorial to go with it!


VPaulV:
Tested, works great

PR #31, opened 2017-07-02
Added import and export commands

adrianherrera:
Just something I did for "fun" over the weekend.

Allows the user to export projects to share with others (e.g. to upload to the s2e-dev google group, share with other S2E users, etc.). When exporting a project, all references to the S2E environment path are replaced with a placeholder that gets rewritten with the new S2E environment path when imported.

I wanted to add a `--contact-vitaly` option to the `s2e export_project` command for when people need help, but maybe that wait for a future PR 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,11 @@ A typical workflow is therefore:
 7. After your analysis has finished, a number of subcommands exist to analyze
    and summarize your results, e.g. the ``coverage`` subcommand, etc.
 
-The `s2e info` command can be used to display a summary of the S2E environment.
-To grab the latest changes from the git repositories, run `s2e update`.
+Other useful commands:
+
+* `s2e info` can be used to display a summary of the S2E environment.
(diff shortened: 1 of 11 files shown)


vitalych:
This is a very nice idea! Does it also include all the symlinked binary files?

adrianherrera:
The target binary is included. Seeds are recipes are also exported if they exist. However, the guestfs and guest tools are not exported - these symlinks get reconstructed on import.

vitalych:
Ideally, import/export should rebuild an identical setup to make it easier to reproduce errors. Unfortunately, this may also mean huge archive. I don't know how to solve this problem easily, but the current approach may be good enough in practice. We'll have to see.

adrianherrera:
Yep, I agree. It was a trade-off, and in the end I decided to go with a smaller archive. Let's see how it goes!

PR #35, opened 2017-07-07
coverage: fixed unable to process null include_directory file

gkso:
I used a simple program to test "s2e coverage lcov", and it didn't generate any coverage report. I found the original script skip any compilation unit without "include_directory" property. So I removed that branch, I'm not sure if it's the correct way to do that. Please check with it. Thanks!

### DIFF
--- a/s2e_env/commands/code_coverage/lcov.py
+++ b/s2e_env/commands/code_coverage/lcov.py
@@ -77,38 +77,37 @@ def _get_file_line_coverage(target_path, addr_counts):
             line_program = dwarf_info.line_program_for_CU(cu)
             cu_filepath = cu.get_top_DIE().get_full_path()
 
-            if line_program['include_directory']:
-                # The actual source path depends on the DWARF line info state
-                # machine
-                src_path = cu_filepath
-                src_dir = os.path.dirname(src_path)
-                for entry in line_program.get_entries():
-                    state = entry.state
+            # Set the default dir and file path
+            src_path = cu_filepath
+            src_dir = os.path.dirname(src_path)
 
-                    if state is None:
-                        # Special handling for commands that don't set a new
-                        # state
-                        if entry.command == dwarf_consts.DW_LNS_set_file:
-                            file_entry = line_program['file_entry'][entry.args[0] - 1]
-                            if file_entry.dir_index == 0:
-                                # Current directory
-                                src_path = os.path.join(src_dir, file_entry.name)
-                            else:
-                                include_dir = line_program['include_directory']
-                                src_path = os.path.join(src_dir, include_dir[file_entry.dir_index - 1],
-                                                        file_entry.name)
-                        elif entry.command == dwarf_consts.DW_LNE_define_file:
+            for entry in line_program.get_entries():
+                state = entry.state
+
+                if state is None:
+                    # Special handling for commands that don't set a new
+                    # state
+                    if entry.command == dwarf_consts.DW_LNS_set_file:
+                        file_entry = line_program['file_entry'][entry.args[0] - 1]
+                        if file_entry.dir_index == 0:
+                            # Current directory
+                            src_path = os.path.join(src_dir, file_entry.name)
+                        elif line_program['include_directory']:
                             include_dir = line_program['include_directory']
-                            src_path = os.path.join(src_dir, include_dir[entry.args[0].dir_index])
-                    elif not state.end_sequence:
-                        # If this address is one that we executed in S2E, save
-                        # the number of times that it was executed into the
-                        # dictionary. Otherwise set it to 0
-                        if src_path not in file_line_info:
-                            src_path = os.path.realpath(src_path)
-                            file_line_info[src_path] = {}
+                            src_path = os.path.join(src_dir, include_dir[file_entry.dir_index - 1],
+                                                    file_entry.name)
+                    elif entry.command == dwarf_consts.DW_LNE_define_file and line_program['include_directory']:
+                        include_dir = line_program['include_directory']
+                        src_path = os.path.join(src_dir, include_dir[entry.args[0].dir_index])
+                elif not state.end_sequence:
+                    # If this address is one that we executed in S2E, save
+                    # the number of times that it was executed into the
+                    # dictionary. Otherwise set it to 0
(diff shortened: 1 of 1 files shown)


adrianherrera:
Hmm interesting, I can't really remember why I added that check. I'm sure it was because https://github.com/eliben/pyelftools/blob/master/scripts/readelf.py did it too, but I just had a look and that doesn't seem to be the case.

Did you do the cat tutorial too? Do your changes work with it as well?

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -77,38 +77,37 @@ def _get_file_line_coverage(target_path, addr_counts):
...
+                if state is None:
+                    # Special handling for commands that don't set a new
+                    # state
+                    if entry.command == dwarf_consts.DW_LNS_set_file:
+                        file_entry = line_program['file_entry'][entry.args[0] - 1]
+                        if file_entry.dir_index == 0:
+                            # Current directory
+                            src_path = os.path.join(src_dir, file_entry.name)
+                        else:
                             include_dir = line_program['include_directory']
```
adrianherrera: What will happen here now that you've removed the `include_directory` check?

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -77,38 +77,37 @@ def _get_file_line_coverage(target_path, addr_counts):
...
-                        # dictionary. Otherwise set it to 0
-                        if src_path not in file_line_info:
-                            src_path = os.path.realpath(src_path)
-                            file_line_info[src_path] = {}
-
-                        file_line_info[src_path][state.line] = addr_counts.get(state.address, 0)
+                            src_path = os.path.join(src_dir, include_dir[file_entry.dir_index - 1],
+                                                    file_entry.name)
+                    elif entry.command == dwarf_consts.DW_LNE_define_file:
+                        include_dir = line_program['include_directory']
```
adrianherrera: Here is another use of `include_directory` that may not exist now that the check has been removed.

adrianherrera: Actually, these two lines are probably why the check is there :)

s2e_env/commands/code_coverage/lcov.py
gkso: Yes, here it should have a check on include_directory, when dir_index is non-zero, there must be include_directory, otherwise it's ill-format

s2e_env/commands/code_coverage/lcov.py
gkso: Yes, it also requires a check here. Since include_directory is optional, in my case it doesn't appear, so i guess the program has also need to process that case.

gkso:
I just tried the cat tutorial, it looks fine... I tried to minimize the modification. The only change is that no matter include_directory exist or not, it will execute the following code

adrianherrera:
Can you please add the check for `line_program[include_directory]` in those two parts of the code indicated above? Then I think this should be okay to merge.

gkso:
Hi Adrian, I've added the two checks. Please check with it. Thanks.

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -77,38 +77,43 @@ def _get_file_line_coverage(target_path, addr_counts):
...
+            src_dir = os.path.dirname(src_path)
+
+            for entry in line_program.get_entries():
+                state = entry.state
+
+                if state is None:
+                    # Special handling for commands that don't set a new
+                    # state
+                    if entry.command == dwarf_consts.DW_LNS_set_file:
+                        print(line_program['file_entry'])
```
adrianherrera: Please remove these print statements

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -77,38 +77,43 @@ def _get_file_line_coverage(target_path, addr_counts):
...
+                        print(entry.args)
+                        file_entry = line_program['file_entry'][entry.args[0] - 1]
+                        if file_entry.dir_index == 0:
+                            # Current directory
+                            src_path = os.path.join(src_dir, file_entry.name)
+                        else:
```
adrianherrera: Do you think it is necessary to throw an exception here? Maybe we can just do this:

```python
elif line_program['include_directory']:
    include_dir = line_program['include_directory']
    src_path = os.path.join(src_dir, include_dir[entry.args[0].dir_index])
```

In the original code if the check for the `include_directory` failed then the function would just return. So I think this is fine

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -77,38 +77,43 @@ def _get_file_line_coverage(target_path, addr_counts):
...
-                        # the number of times that it was executed into the
-                        # dictionary. Otherwise set it to 0
-                        if src_path not in file_line_info:
-                            src_path = os.path.realpath(src_path)
-                            file_line_info[src_path] = {}
-
-                        file_line_info[src_path][state.line] = addr_counts.get(state.address, 0)
+                            src_path = os.path.join(src_dir, include_dir[file_entry.dir_index - 1],
+                                                    file_entry.name)
+                    elif entry.command == dwarf_consts.DW_LNE_define_file:
```
adrianherrera: Likewise here we could just have

```python
elif entry.command == dwarf_consts.DW_LNE_define_file and line_program['include_directory']:
    include_dir = line_program['include_directory']
    src_path = os.path.join(src_dir, include_dir[entry.args[0].dir_index])
```

gkso: Sure. agree. let's do that.

gkso:
Done.Thanks

adrianherrera:
Perfect! Please squash into a single commit and then I will merge it.

Many thanks!

gkso:
I tried to squash them into one commit, there is also another merge commit. does it look right?

adrianherrera:
You can squash that one too :)

gkso:
Sorry for making it hard for you. I just reverted the changes, merged with upstream first and then resubmitted my change...

adrianherrera:
No problem, there is no need to apologize! Thanks for the PR!

PR #47, opened 2017-07-28
bootstrap: start the LinuxMonitor driver

adrianherrera:
Goes with https://github.com/S2E/s2e-linux-kernel/pull/4


s2e_env/templates/bootstrap.windows_dll.sh
```diff
@@ -21,8 +21,10 @@ function target_init {
     PREFIX=
     {% endif %}
 
+    # Start the WindowsMonitor driver
```
vitalych: Avoid code duplication with bootstrap.windows.sh

adrianherrera: Code point. Refactored and removed.

PR #54, opened 2017-09-09
Added Binary Ninja disassembler backend

adrianherrera:
Last disassembler backend, I swear!! I just really like [Binary Ninja](https://binary.ninja), and I don't really have access to IDA anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,13 @@ We assume that you are working on an Ubuntu 14.04 or 16.04 64-bit OS.
 [Repo](https://code.google.com/p/git-repo/) only works with Python 2.7, so you
 should use Python 2.7 too. You will also need `gcc` and `python-dev` installed.
 
-Some commands (such as basic block coverage) require either
-[IDA Pro](https://www.hex-rays.com/products/ida/) or [Radare](https://rada.re/)
-for performing static analysis. If you are using Radare, you will also have to
-install `r2pipe` via pip.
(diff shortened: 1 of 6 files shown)


vitalych:
I think it would make sense at this point to add protobuf-based cfg support to s2e, to abstract away various disassembly engines. Each engine produces a protobuf cfg consumed by various s2e plugins. Otherwise, each plugin would have to support N disassemblers.

adrianherrera:
Sorry, I don't understand. This has nothing to do with S2E plugins. This is so that I can use s2e-env to generate and view basic block coverage in Binary Ninja.

vitalych:
I meant that we should get output of all 3rd party tools in a generic format (e.g., protobuf), so that s2e-env (and s2e) can easily interface with them. Right now if you need coverage from IDA and Binary Ninja, you would write 2 different modules that do the same thing. Instead, if IDA and Binary Ninja generated protobuf format, you would write your coverage tool only once.

Unless what you have in mind is a Binary Ninja plugin that would display coverage directly in Binary Ninja iteself, in which case my remark is irrelevant.

adrianherrera:
We already have a generic format, it's the translation block trace in JSON format produced by S2E. Each third party tool still needs to convert this translation block JSON into basic blocks based on how its disassembly engine and then render this information graphically.

I'm not sure what adding protobuf gives on top of the translation block JSON.

PR #56, opened 2017-10-24
execution_trace: updated TraceProcessUnload struct

adrianherrera:
Now includes the process return code. Goes with https://github.com/S2E/libs2eplugins/pull/18


adrianherrera:
Hi @vitalych,

Could you please review this PR also. I'll have to merge all the `improvement/include-returncode-process-unload` branches together, otherwise things will break.

Thanks!

PR #88, opened 2018-01-10
Issue/84 drivers

vitalych:
Depends on prior PR, look at last 4 commits


s2e_env/infparser/device.py
```diff
@@ -0,0 +1,129 @@
...
+        return ret
+
+    def get_s2e_cfg(self):
+        pass
+
+    def is_pci(self):
+        pass
+
+    def __unicode__(self):
+        return 'DEVICE %s %s [%s]' % (self.install_section, self.hardware_id, self.name)
```
adrianherrera: Should this string literal be prefixed by a `u` (since it's unicode)?

s2e_env/infparser/CaseInsensitiveDict.py
```diff
@@ -0,0 +1,97 @@
+"""
```
adrianherrera: The filename should be `case_insensitive_dict.py`

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+
+import logging
+import re
+
+from .CaseInsensitiveDict import CaseInsensitiveDict
```
adrianherrera: `from .case_insensitive_dict`

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+from .device import Device, InstallInfo
+from .parser import InfFileParser
+
+logger = logging.getLogger('infparser')
+
+_macroPattern = re.compile('(%.+%)')
```
adrianherrera: If this is a constant it should be capitalised (e.g. `MACRO_PATTERN`).

The regex string literal should probably be prefixed by an `r` (not really necessary here, but it's still good practice).

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+            tokens_found.append(m.group(1))
+
+        for m in tokens_found:
+            expanded = self.get_string(m)
+            key = key.replace(m, expanded)
+
+        return key
+
+    def has_manufacturers(self):
+        if not self._sections.has_key('manufacturer'):
```
adrianherrera: Can this function be simplified to:

```py
return self._sections.has_key('manufacturer')
```

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # pylint: disable=protected-access
+        ret._sections = InfFileParser.parse(filename)
+        return ret
+
+    @staticmethod
+    def from_string(input_string):
+        ret = InfFile('noname.inf')
+
+        # pylint: disable=protected-access
```
adrianherrera: This seems like a code smell to me. If you need to disable the pylint warning, why not just make `_sections` "public" (i.e. without the leading underscore?) Or you should be using Python property decorators (https://docs.python.org/2.7/library/functions.html#property) - e.g. having a `sections.setter` decorator on a `sections` method.

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+    def _is_string_key(string_spec):
+        return string_spec[0] == '%' and string_spec[len(string_spec) - 1] == '%'
+
+    @staticmethod
+    def to_string_key(string_spec):
+        k = string_spec.split('%')
+        return k[1]
+
+    def get_string(self, key):
+        if not self._is_string_key(key):
```
adrianherrera: If the key is not a string, then the key is returned? That seems... odd for a function named `get_string` (because it may not return a string). Does it make more sense to return `None`?

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+    # [models-section-name.TargetOSVersion]  (Windows XP and later versions of Windows)
+    # device-description=install-section-name[,hw-id][,compatible-id...]
+    # [device-description=install-section-name[,hw-id][,compatible-id]...] ...
+    #
+    # Returns a mapping from the device key to the actual device
+    def get_models(self, dev_key, version):
+        ret = CaseInsensitiveDict()
+
+        if version is not None:
+            dev_key = dev_key + '.' + version
```
adrianherrera: `dev_key = '%s.%s' % (dev_key, version)` :)

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+    # [models-section-name] |
+    # [models-section-name.TargetOSVersion]  (Windows XP and later versions of Windows)
+    # device-description=install-section-name[,hw-id][,compatible-id...]
+    # [device-description=install-section-name[,hw-id][,compatible-id]...] ...
+    #
+    # Returns a mapping from the device key to the actual device
+    def get_models(self, dev_key, version):
+        ret = CaseInsensitiveDict()
+
+        if version is not None:
```
adrianherrera: This can be simplified to:

```py
if version:
```

s2e_env/infparser/file.py
```diff
@@ -0,0 +1,247 @@
...
+            logger.warn('Section %s does not exist', install_section_key)
+            return ret
+
+        for fk in self._sections.prefixed_keys(install_section_key):
+            # Get the files to copy
+            install_section = self._sections[fk].data
+            install_info = InstallInfo()
+
+            # Remove the potential suffix
+            for ver in ['.nt', '.ntx86', '.ntia64', '.ntamd64']:
```
adrianherrera: It's generally considered more "Pythonic" to make this a tuple (because tuples are immutable and this is a constant).

s2e_env/infparser/parser.py
```diff
@@ -0,0 +1,131 @@
...
+import codecs
+import logging
+import re
+
+from .CaseInsensitiveDict import CaseInsensitiveDict
+from .section import InfSection
+
+logger = logging.getLogger('infparser')
+
+sectionPattern = re.compile(r'\[(.+)\]')
```
adrianherrera: `SECTION_PATTERN`

s2e_env/infparser/parser.py
```diff
@@ -0,0 +1,131 @@
...
+
+class InfFileParser(object):
+    def __init__(self):
+        self._sections = CaseInsensitiveDict()
+
+    @staticmethod
+    def _read_file(filename):
+        decoded = False
+
+        for codec in ['utf-16', 'utf-8']:
```
adrianherrera: Make this a tuple

s2e_env/infparser/parser.py
```diff
@@ -0,0 +1,131 @@
...
+                    sections[current_section.name] = current_section
+                current_section = InfSection(section_name)
+                continue
+
+            # Skip
+            if not current_section.name:
+                continue
+
+            # The line is of the form key=value
+            key_value = line.split("=")
```
adrianherrera: single quotes

s2e_env/infparser/version.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if version is None:
+            return
+
+        version = unicode(version.lower())
+        if not version.startswith("nt"):
+            return
+
+        # Fields can be empty...
+        v = version.split(".")
```
adrianherrera: single quotes

s2e_env/infparser/version.py
```diff
@@ -0,0 +1,96 @@
...
+            self.product_type = int(v[3], 16) if v[3] else None
+        if len(v) >= 5:
+            self.suite_mask = int(v[4], 16) if v[4] else None
+
+        # Normalize the arch
+        if self.architecture == 'ntx86':
+            self.architecture = 'nt'
+
+    def matches(self, desired_version):
+        if self.architecture is not None and desired_version.architecture is not None:
```
adrianherrera: Can just be simplified to

```py
if self.architecture and desired_version.architecure:
```

s2e_env/infparser/version.py
```diff
@@ -0,0 +1,96 @@
...
+        if v[1] is None or dv[1] is None:
+            if v[0] is not None and dv[0] is not None:
+                if v[0] < dv[0]:
+                    return False
+
+        return True
+
+    def __str__(self):
+        if self.architecture is None:
+            return "Windows 2000 and later"
```
adrianherrera: Single quotes

s2e_env/commands/new_project.py
```diff
@@ -119,30 +195,21 @@ def handle(self, *args, **options):
...
-
-            # If we find a match, create that project. The user instructions
-            # are returned
-            if matches:
-                options['target'] = target_path
-                options['target_arch'] = arch
-
-                return call_command(Project(proj_config_class), **options)
+        arch, proj_config_class = get_arch(target_path)
+        if arch is not None:
```
adrianherrera: `if arch:`

s2e_env/infparser/device.py
```diff
@@ -0,0 +1,129 @@
...
+        elif 'USB' in enumerator[0]:
+            ret = USBDevice(name, install_section, hardware_id)
+        else:
+            ret = Device(name, install_section, hardware_id)
+
+        ret.version = version
+        return ret
+
+    def get_s2e_cfg(self):
+        pass
```
adrianherrera: I don't really like having `pass` for abstract methods - this has the potential for breaking things if a mistake is made when subclassing (blame Python's "type system"). I'd prefer that you either throw an exception or return a "default" value.

s2e_env/commands/new_project.py
```diff
@@ -119,30 +195,21 @@ def handle(self, *args, **options):
...
+            options['target'] = target_path
+            options['target_files'] = [target_path]
+            options['target_arch'] = arch
+            options['modules'] = [(os.path.basename(target_path), False)]
+
+            options['processes'] = []
+            if proj_config_class is not WindowsDLLProjectConfiguration:
```
adrianherrera: It's better to use `isinstance`

s2e_env/commands/code_coverage/basic_block.py
```diff
@@ -86,43 +86,49 @@ class BasicBlockCoverage(ProjectCommand):
               'Covered basic blocks: {num_covered_bbs} ({percent:.1%})'
```
adrianherrera: This commit is cool! Let's get this stuff merged first, and then I'll rebase against it (because it will definitely cause conflicts with my per-state coverage code).

s2e_env/commands/coverage.py
```diff
@@ -58,16 +62,12 @@ def handle(self, *args, **options):
             # Select the disassembler backend
             disassembler = options.pop('disassembler', ())
             if disassembler == 'ida':
-                from .code_coverage.ida_basic_block import IDABasicBlockCoverage
```
adrianherrera: Hmmm you must be using a new version of pylint? I've never got errors about this - I'll have to check do an upgrade

s2e_env/commands/image_build.py
```diff
@@ -319,7 +319,8 @@ def handle(self, *args, **options):
             image_downloader = ImageDownloader(templates)
             image_downloader.download_images(image_names, self.image_path())
 
-            return 'Successfully downloaded images: %s' % ', '.join(image_names)
+            logger.info('Successfully downloaded images: %s', ', '.join(image_names))
```
adrianherrera: Why did you do this? All `Command` `handle` methods return a string on success.

s2e_env/commands/image_build.py
```diff
@@ -349,7 +350,8 @@ def handle(self, *args, **options):
 
         self._invoke_make(img_build_dir, rule_names, num_cores, options['iso_dir'])
 
-        return 'Built image \'%s\'' % image_name
+        logger.info('Built image \'%s\'', image_name)
```
adrianherrera: Why?

s2e_env/commands/import_project.py
```diff
@@ -125,7 +125,7 @@ def _decompress_archive(self, archive_path):
                 directory=self.projects_path(), _fg=True, _out=sys.stdout,
                 _err=sys.stderr)
         except ErrorReturnCode as e:
-            raise CommandError('Failed to decompress project archive - %s', e)
+            raise CommandError('Failed to decompress project archive', e)
```
adrianherrera: There is definitely a mistake here. But I believe the fix should be:

```py
'Failed to decompress project archive - %s' % e
```

Does the `CommandError` class (which is basically just `Exception`) provide a constructor that takes two arguments?

s2e_env/infparser/parseinf.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python
```
adrianherrera: Why is this here? If it's in s2e-env, I think it should be an s2e-env command. Otherwise I don't think it should be here. How would you even run this if you did a `pip install`?

s2e_env/infparser/file.py
vitalych: Te error makes no sense to me. The field is accessed from a static class member, not be someone else.

s2e_env/infparser/file.py
vitalych: Yeah, this is INF format issue. If the INF file is malformed, it's better to return the raw key.
I'll try to comment all these functions and explain the format a bit more.

s2e_env/infparser/file.py
vitalych: ok, though this rule makes no sense to me

s2e_env/commands/coverage.py
vitalych: Not really, I just thought it would be nicer to put these at the beginning of the file.

s2e_env/commands/image_build.py
vitalych: Because pylint complained that we have a mix of returns that return nothing / something.
I didn't feel like putting an artificial string on every return, so I chose to return nothing everywhere.
I also find it strange to return something to indicate success, given the language has exceptions.

s2e_env/infparser/device.py
vitalych: yes

s2e_env/infparser/CaseInsensitiveDict.py
vitalych: ok

s2e_env/infparser/file.py
vitalych: ok

s2e_env/infparser/file.py
vitalych: ok

s2e_env/infparser/file.py
vitalych: yes

s2e_env/infparser/file.py
vitalych: ok

s2e_env/infparser/file.py
vitalych: ok

s2e_env/infparser/parser.py
vitalych: ok

s2e_env/infparser/parser.py
vitalych: ok

s2e_env/infparser/parser.py
vitalych: ok

s2e_env/infparser/version.py
vitalych: ok

s2e_env/commands/image_build.py
adrianherrera: Your arguments are reasonable and have convinced me :) However, let's just make sure that this is consistent

s2e_env/infparser/version.py
vitalych: ok

s2e_env/infparser/version.py
vitalych: ok

s2e_env/commands/new_project.py
vitalych: ok

s2e_env/infparser/device.py
vitalych: ok

s2e_env/commands/new_project.py
vitalych: ok

s2e_env/infparser/parseinf.py
vitalych: removed.
I run this with python -m anyway.

s2e_env/infparser/file.py
adrianherrera: Right, I see. Bizarre!

s2e_env/infparser/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""
```
adrianherrera: Is it necessary to have this copyright here (in an empty __init__ file)? AFAIK we don't do this for any other __init__ file.

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -47,21 +47,27 @@ class LineCoverage(ProjectCommand):
...
-        if options.get('html', False):
-            lcov_html_dir = self._gen_html(lcov_info_path)
-            return 'Line coverage saved to %s. An HTML report is available in %s' % (lcov_info_path, lcov_html_dir)
+            if options.get('html', False):
+                lcov_html_dir = self._gen_html(module, lcov_info_path)
+                logger.info('Line coverage saved to %s. An HTML report is available in %s',
+                            lcov_info_path, lcov_html_dir)
 
-        return 'Line coverage saved to %s' % lcov_info_path
+            logger.info('Line coverage saved to %s', lcov_info_path)
```
adrianherrera: This is still not consistent, see https://github.com/S2E/s2e-env/pull/88/commits/11efb44715a00863c15f63ebe7e5ee0ed75b6810#diff-57a06c052391c16158efd88597edc051R120

s2e_env/infparser/parseinf.py
adrianherrera: Sorry sorry sorry, I was not clear at all! I wasn't referring to the shebang (which should be there for scripts), but the entire script in general.

Why is it not an s2e-env command? How do I run this from `s2e` if I've done a `pip install`? I can't run it by doing `s2e parseinf ...`, so I have to do a `python -m /path/to/my/s2e-env/source/code`? The script is not externally exposed in anyway,.

I think it should either be in the tools repo, or it should be an `s2e` command.

s2e_env/commands/coverage.py
adrianherrera: Fair enough.

s2e_env/infparser/__init__.py
vitalych: There are plenty of __init__ files for which we do. I prefer having a copyright header in every source file, it removes all ambiguity.

s2e_env/commands/code_coverage/lcov.py
vitalych: I don't understand what's not consistent

adrianherrera: Sorry, I'm not being clear (again!). I mean that all the other `Command`s return a string. If we're going to make this change (and I agree with the points that you raised), I'd rather we do all the `Command.handle` methods in one commit. I think that this is preferable than leaving it in this in-between state where lcov's `Command.handle` has been changed but the rest have just be left as-is (and may very well be forgotten).

I'd rather this be reverted in this commit, and we can make a new PR later that changes all the commands in one go.

adrianherrera:
Sorry, I know I'm being a pain and nitpicking this logger thing, but it's just a pet peeve of mine when part of a codebase gets "refactored" and the rest is "left to do another time". I'm hyper-conscious of falling into this trap, so I'd prefer we do this change (even though it is super minor!) in one go. It will ease my OCD :)

adrianherrera:
Looks good now!!! A few things to fix:

* Use `logger.success` instead of `logger.info`, as per https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/command.py#L186
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/command.py#L186 - replace the code on lines 186 - 189 with a single call to `self.handle(*args, **options)` - no need to check for return strings now
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/new_project.py#L213 - remove the return
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/new_project.py#L216 - remove the return
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/new_project.py#L152 - remove the return
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/project_creation/__init__.py#L178 - change this return to `logger.success(_create_instructions(config))`
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/image_build.py#L363 - remove the redundant return
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/update.py#L49 - should be `logger.success`
* https://github.com/S2E/s2e-env/blob/issue/84-drivers/s2e_env/commands/update.py#L71 - should be `logger.info` :)

This should be super quick to do, and I'll approve it as soon as it's done (promise 🤞)

adrianherrera:
Maybe squash the last 2 commits - i.e. squash "Use logger.success where appropriate" into "commands: don't return string from handle()". I don't think there's any reason to keep "Use logger.success where appropriate" separate.

Issue #68, opened 2017-11-21
Build guest images without KVM


adrianherrera:
KVM should be optional when building guest images - it is only needed to improve performance. Some systems may not have KVM available - e.g. the Windows Subsystem for Linux (WSL) - but guest images should still be build-able on these systems.

adrianherrera:
Implemented. Documented in https://github.com/S2E/docs/pull/31

PR #95, opened 2018-01-18
image_build: added option to disable KVM

adrianherrera:
This allows images to be built on WSL, etc. Addresses https://github.com/S2E/s2e-env/issues/68

### DIFF
--- a/s2e_env/commands/image_build.py
+++ b/s2e_env/commands/image_build.py
@@ -119,9 +119,9 @@ def _check_kvm():
     This is required by libs2e to communicate with QEMU.
     """
     if not os.path.exists(os.path.join(os.sep, 'dev', 'kvm')):
-        raise CommandError('KVM is required to build images. Check that '
-                           '/dev/kvm exists. Alternatively, you can also '
-                           'download pre-built images (-d option).')
+        raise CommandError('KVM interface not found - check that /dev/kvm '
+                           'exists. Alternatively, you can disable KVM (-n '
+                           'option) or download pre-built images (-d option)')
 
 
 def _check_vmlinux():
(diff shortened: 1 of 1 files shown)


s2e_env/commands/image_build.py
```diff
@@ -301,12 +308,12 @@ def handle(self, *args, **options):
 
         if options['clean']:
             self._invoke_make(img_build_dir, ['clean'], num_cores)
-            return
+            return None
```
vitalych: why?

s2e_env/commands/image_build.py
```diff
@@ -338,7 +345,9 @@ def handle(self, *args, **options):
         _check_product_keys(templates, image_names)
         _check_iso(templates, options['iso_dir'], image_names)
 
-        _check_kvm()
+        if self._use_kvm:
+            _check_kvm()
```
vitalych: Does this require fixing guest-images too?

s2e_env/commands/image_build.py
adrianherrera: Hahah honestly I have no idea why I did this :\ Thanks for pointing it out

s2e_env/commands/image_build.py
adrianherrera: Hmmm I don't think so? I just "clear" the `QEMU_KVM` variable from [here](https://github.com/S2E/guest-images/blob/master/Makefile.common#L35)

adrianherrera: Or am I missing something (or misunderstanding!)

vitalych: No, I just forgot how this options was passed, this looks good.

vitalych:
I can't wait to see if this will make image building faster than using nested virtualization.

adrianherrera:
Changes made.

FWIW I don't have a Windows 10 box to test WSL. I just tested on my regular Ubuntu 16.04 box, and while it was substantially slower, it still built an image successfully.

PR #108, opened 2018-02-25
Empty project

vitalych:
This is useful when there isn't really a target to analyze, or the desired target is not supported by s2e-env. This gives the option of configuring everything manually. The command still provides a default project skeleton for the type of OS requested.


s2e_env/commands/new_project.py
```diff
@@ -151,6 +161,66 @@ def _handle_inf(target_path, **options):
...
+        raise CommandError('%s is not a valid target for S2E analysis' %
+                           target_path)
+
+
+def _handle_empty_project(**options):
+    if not options['no_target']:
+        raise CommandError('No target binary specified. Use the -nt option to create an empty project')
+
+    if not options['image']:
+        raise CommandError('An empty project must specify the image to use')
```
adrianherrera: Maybe keep these messages consistent and print the flag that is required to specify and image.

vitalych: ok

PR #119, opened 2018-03-27
trace_entries: improved test case format

vitalych:
Signed-off-by: Vitaly Chipounov <vitaly@cyberhaven.io>

### DIFF
--- a/s2e_env/execution_trace/trace_entries.py
+++ b/s2e_env/execution_trace/trace_entries.py
@@ -22,6 +22,7 @@ SOFTWARE.
 
 import binascii
 import logging
+import re
 import struct
 
 from enum import Enum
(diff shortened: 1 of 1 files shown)


s2e_env/execution_trace/trace_entries.py
```diff
@@ -769,14 +782,38 @@ def _initialize_test_case_items(self):
...
-        return self._testcase
+        return {'testcase': self._testcase}
 
     def as_json_dict(self):
-        ret = {}
+        ret = []
 
-        for k, v in self._testcase.iteritems():
-            ret[k] = binascii.hexlify(v)
+        for entry in self._testcase:
```
adrianherrera: Probably nicer to extract the tuple elements, i.e.

```py
for var_name, v in self._testcase:
    tc = (var_name, binascii.hexlify(v))
    ret.append(tc)
```

Although now that I look at it, you could just as easily do

```py
ret = [(var_name, binascii.hexlify(v)) for var_name, v in self._testcase]

return {'testcase': ret}

PR #139, opened 2018-05-26
Issue/128 povgen

vitalych:


s2e_env/commands/forkprofile.py
```diff
@@ -75,7 +75,8 @@ def get_profile(self):
             for rel_pc, count in counts.iteritems():
                 try:
                     sym, fcn = self._syms.get(mod.path, rel_pc)
-                except Exception:
+                except Exception as e:
+                    logger.debug(e, exc_info=1)
```
adrianherrera: Should these be going to a different stream? E.g. `warning`

adrianherrera:
Looks like this branch needs to be rebased with master

s2e_env/commands/forkprofile.py
vitalych: No, the exception is normal, it would make it too verbose.

adrianherrera: ACK

adrianherrera:
Looks like it needs a rebase

PR #158, opened 2018-06-09
Issue/157 sympath

vitalych:


s2e_env/commands/new_project.py
```diff
@@ -115,8 +115,36 @@ def get_arch(target_path):
...
 
+def _gen_win_driver_project(target_path, file_paths, **options):
+    first_sys_file = None
+    for f in file_paths:
+        if f.endswith('.sys'):
+            first_sys_file = f
+
+    # TODO: prompt the user to select the right driver
+    if not first_sys_file:
+        raise CommandError('Could not find any *.sys file')
```
adrianherrera: I don’t really know much about windows drivers, but can this error message be expanded in some way to help the user? Maybe by either saying why a sys file is needed or where it can be found? I don’t know if this makes sense, but I just feel that the error message could be slightly more descriptive

s2e_env/commands/code_coverage/lcov.py
```diff
@@ -35,6 +36,116 @@
...
+                        ``_get_file_line_coverage``.
+
+    Returns:
+        The file path where the line coverage information was written to.
+    """
+    logger.info('Writing line coverage to %s', lcov_path)
+
+    with open(lcov_path, 'w') as f:
+        f.write('TN:\n')
+        for src_file in file_line_info.keys():
```
adrianherrera: When iterating over dictionary keys, you don't need `.keys()`, you can just iterate over `file_line_info`. This is considered more "Pythonic" :)

adrianherrera:
One minor style fix and then I'll approve.

s2e_env/commands/code_coverage/lcov.py
vitalych: fixed. I wonder why pylint doesn't catch it.

PR #181, opened 2018-06-30
init: improved reusing existing S2E install

adrianherrera:
* Sanity checks on the existing S2E installation
* Improved user messages


vitalych:
Is there some documentation about it? I didn't know about this script until now.

adrianherrera:
Yeah it's in the README (although I now realised it's outdated. I'll update it)

vitalych:
Thanks! Perhaps also adding a doc header to the script would help.

PR #182, opened 2018-06-30
commands: added single path(sp) support

humeafo:


s2e_env/commands/run.py
```diff
@@ -182,6 +182,8 @@ def add_arguments(self, parser):
                             help='Optional arguments to the S2E launcher script')
         parser.add_argument('-c', '--cores', required=False, default=1,
                             type=int, help='Number of cores to run S2E on')
+        parser.add_argument('-s', '--single_path', action='store_true',
```
adrianherrera: Please change to `--single-path`

s2e_env/commands/run.py
```diff
@@ -262,11 +265,15 @@ def handle(self, *args, **options):
                 qmp_server.shutdown()
                 qmp_server.server_close()
 
-    def _setup_env(self, project_args, cores, qmp_socket):
+    def _setup_env(self, project_args, cores, qmp_socket, single_path):
+        if single_path and cores > 1:
+            logger.warning("S2E's multi-processes support is not available in single path mode")
```
adrianherrera: Our string style is for single quotes :)

s2e_env/commands/run.py
```diff
@@ -262,11 +265,15 @@ def handle(self, *args, **options):
...
+
         qemu_build = self._project_desc['image']['qemu_build']
         qemu = self.install_path('bin',
                                  'qemu-system-%s' % qemu_build)
-        libs2e = self.install_path('share', 'libs2e', 'libs2e-%s-s2e.so' % qemu_build)
+        s2ename = 'libs2e-%s-s2e%s.so' % (qemu_build, '_sp' if single_path else '')
```
adrianherrera: Hmmm might make sense to rename to this instead:

* `s2ename` -> `libs2e`
* `libs2e` -> `libs2e_path`

s2e_env/templates/launch-s2e.sh
```diff
@@ -10,10 +10,16 @@ ENV_DIR="{{ env_dir }}"
 INSTALL_DIR="$ENV_DIR/install"
 BUILD_DIR="$ENV_DIR/build/s2e"
 BUILD=debug
+S2E_SP_SUFFIX=
 
 # Comment this out to enable QEMU GUI
 GRAPHICS=-nographic
 
+if [ "x$1" = "xs2esp" ]; then
```
adrianherrera: Maybe make the option `s2e_sp` (so the check becomes `xs2e_sp`)

vitalych:
Nice addition! Could you also clean up the bootstrap scripts and s2e-config templates?
You'll need to remove test case generators, searchers, etc. Just keep the basic set of plugins that don't touch symbolic data (BaseInstructions, HostFiles, OSMonitors, etc...). Use an {%if..%} block for that.

humeafo:
@vitalych only basic s2e-config and bootstrap.sh changes are made, I'm not sure whether some specific targets need adjustments too, maybe a tutorial can be added later.

@adrianherrera all done.

s2e_env/templates/s2e-config.lua
```diff
@@ -153,6 +153,7 @@ pluginsConfig.ModuleExecutionDetector = {
     {% endfor %}
 }
 
+{% if single_path_mode == false %}
```
vitalych: I think it's better to not use the single_path_mode mode variable inside s2e-config.lua. Instead, use a flag for each plugin (e.g., `if use_fork_profiler`). Then the set of plugins can be configured in the commands code. It will be easier to handle different combinations this way.

s2e_env/commands/project_creation/__init__.py
```diff
@@ -267,6 +275,7 @@ def _create_bootstrap(self, config):
             'target_files': config['target_files'],
             'modules': config['modules'],
             'processes': config['processes'],
+            'single_path_mode': config['single_path_mode'],
```
vitalych: don't pass this var here to config files (see the other comment)

s2e_env/commands/new_project.py
```diff
@@ -269,6 +269,9 @@ def add_arguments(self, parser):
         parser.add_argument('--enable-pov-generation', action='store_true', default=False,
                             help='Enables PoV generation')
 
+        parser.add_argument('--single-path-mode', action='store_true', default=False,
+                            help='Enable single path mode')
```
vitalych: The help message could be more descriptive (i.e., beyond repeating the argument itself). I know, the other ones are not very good either (e.g., enable-pov-generation).

s2e_env/commands/run.py
```diff
@@ -262,18 +265,22 @@ def handle(self, *args, **options):
                 qmp_server.shutdown()
                 qmp_server.server_close()
 
-    def _setup_env(self, project_args, cores, qmp_socket):
+    def _setup_env(self, project_args, cores, qmp_socket, single_path):
+        if single_path and cores > 1:
+            logger.warning('Multi-processes support is not available in single path mode')
```
vitalych: How about throwing a CommandError here?

vitalych:
@humeafo I added some more feedback. Regarding the tutorial, I think it shouldn't be postponed (otherwise in my experience it's never going to be written). It doesn't have to be large. Three paragraphs could be enough. See, e.g., the tutorial about the fork profiler:
http://s2e.systems/docs/src/Tools/ForkProfiler.html

Once you apply the feedback, please rebuild the demo docker image and check that it runs and finds PoVs (it's my default smoke test before merging). I really hope to have it automated soon.

humeafo:
Didn't notice that you have merged in alot more codes, maybe this patch should be rebased...

PR #184, opened 2018-07-02
commands: added single path(sp) support

humeafo:


humeafo:
It's rebased because the recent codes merged in, and a mis-operation (push -f) caused the review history loss.

Issue #191, opened 2018-07-05
Current refactored bootstrap.sh can't generate correct SYMB_FILE name


humeafo:
The problem is that ${S2EGET} ${SEED_FILE}.symranges will output "Waiting for S2E mode..." at
https://github.com/S2E/s2e-env/blob/master/s2e_env/templates/bootstrap.sh#L98

then the SYMB_FILE="$(prepare_inputs ${SEED_FILE})" at https://github.com/S2E/s2e-env/blob/master/s2e_env/templates/bootstrap.sh#L154 will got "Waiting for S2E mode...\n"+"${REAL_SYMB_FILE_NAME}" which is obviously won't work.

vitalych:
I don't understand the problem. Is this in single path mode? s2eget should work the same there. The message you see will only happen if you run a non-s2e build of libs2e.

humeafo:
The problem exists seems after your last commit, if you use seed now at runtime SYMB_FILE will get value "Waiting for S2E mode...\nXXXXXXX" so the seeds mode won't work now.

vitalych:
ok, now I get it, thanks. I didn't realize that there would be the waiting for s2e mode message. I suppose a fix would be to redirect s2eget output to /dev/null. Please send a fix.

PR #192, opened 2018-07-06
templates: fix incorrect SYMB_FILE content when using seeds in bootst…

humeafo:
…rap.sh

### DIFF
--- a/s2e_env/templates/bootstrap.sh
+++ b/s2e_env/templates/bootstrap.sh
@@ -84,7 +84,7 @@ function prepare_inputs {
             exit 1
         fi
     else
-        ${S2EGET} ${SEED_FILE}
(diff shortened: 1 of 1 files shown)


humeafo:
This resolves https://github.com/S2E/s2e-env/issues/191

s2e_env/templates/bootstrap.sh
```diff
@@ -84,7 +84,7 @@ function prepare_inputs {
             exit 1
         fi
     else
-        ${S2EGET} ${SEED_FILE}
+        ${S2EGET} ${SEED_FILE} >/dev/null 2>&1
```
vitalych: I think it should be ok to not redirect stderr.

vitalych:
@humeafo  thanks! please squash and rebase.

PR #193, opened 2018-07-06
commands: don't prevent user environment variables to be passed to s2…

humeafo:
…e instance


humeafo:
this resolves https://github.com/S2E/s2e-env/issues/191

PR #195, opened 2018-07-10
templates: use env to preload libs2e.so

matrizzo:
Using set environment was causing QEMU to wait forever on select() on the machine I'm working on (Ubuntu 16.04) but I'm not sure why since both methods should just load the library. This should be tested on other machines before merging.

Signed-off-by: Matteo Rizzo <matteo.rizzo@epfl.ch>


vitalych:
I have never encountered this issue, neither on 16, 18, or 20. Closing for now.

PR #200, opened 2018-07-24
Added support for Ubuntu 18 04

adrianherrera:
Plus other general tidy ups.

### DIFF
--- a/s2e_env/commands/code_coverage/__init__.py
+++ b/s2e_env/commands/code_coverage/__init__.py
@@ -80,6 +80,7 @@ def _tb_to_uint64(tb):
     size = tb[2]
     return start_addr, end_addr, size
 
+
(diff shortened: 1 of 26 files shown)


s2e_env/commands/code_coverage/__init__.py
```diff
@@ -102,7 +103,7 @@ def aggregate_tb_files_per_state(tb_files):
 
             for module_path, coverage in data.iteritems():
                 states = {}
-                if module_path not in ret.keys():
+                if module_path not in ret:
```
vitalych: Could we force pylint to detect these?

adrianherrera: Not sure. This would require pylint to reason about the variable's type; i.e. if it's a `dict` or some arbitrary class with a `keys()` method.

PR #203, opened 2018-08-02
BB coverage speedups

adrianherrera:
A colleague of mine implemented the following basic block coverage speedup. I have tested and the results are consistent.

Some stats (by running `time s2e coverage basic_block`):

**Original n^2 algorithm**:

Total basic blocks: 1359
Covered basic blocks: 210 (15.5%)

real    1m42.149s
user    1m41.424s
sys     0m0.188s 

**New binary search algorithm**:

Total basic blocks: 1359
Covered basic blocks: 210 (15.5%)

real    0m3.833s
user    0m3.184s
sys     0m0.188s

Huuuuge improvement!


s2e_env/commands/code_coverage/basic_block.py
```diff
@@ -96,6 +96,38 @@ def object_hook(self, d):
...
+    num_bbs = len(bbs)
+    left_idx = 1
+    right_idx = num_bbs - 1
+
+    if tb_start_addr <= bbs[0].end_addr:
+        return 0
+    elif tb_start_addr > bbs[right_idx].end_addr:
+        return num_bbs
+
+    while right_idx - left_idx > 1:
```
vitalych: Is there a generic algorithm for that? I remember using bisect to quickly lookup debug symbols.

adrianherrera: So it turns out you can't easily use `bisect` because you are comparing objects of different types: `BasicBlock` objects vs. addresses (i.e. `int`). Also, the comparison [here](https://github.com/S2E/s2e-env/pull/203/files#diff-57a06c052391c16158efd88597edc051R119) is not the typical binary search comparison.

PR #215, opened 2018-11-03
Fix 's2e build -g', which didn't re-install updated binaries.

gkso:
The 's2e build -g' command builds debug version of s2e but doesn't install it to the destination 'install' folder. I found it invokes 'make all-debug' instead of 'make install-debug'. So I guess that's the reason. Would you please confirm it?

### DIFF
--- a/s2e_env/commands/build.py
+++ b/s2e_env/commands/build.py
@@ -88,7 +88,7 @@ class Command(EnvCommand):
             # Run make
             if options['debug']:
                 logger.info('Building S2E (debug) in %s', build_dir)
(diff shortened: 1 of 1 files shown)


adrianherrera:
Yes, this was done on purpose.

The release version should be in `install/`, and this is the version that should be used by default. The debug version is invoked when the `debug` option is passed to `launch-s2e.sh`. If you installed the debug version to `install/`, then it would always be used by default (even when not passing the `debug` option to `launch-s2e.sh`, which is undesirable).

PR #222, opened 2018-12-02
Improvement/dll-tests

adrianherrera:
New tests for DLLs (both project creation and analysis).

I've included the source code for the DLL + a compiled binary. Not sure if you have a preference to only include the source, and then build the DLL when you run the tests (requires mingw), or you'd just prefer the binary, or happy to have both. Thoughts?


tests/data/src/Myputs.c
```diff
@@ -0,0 +1,53 @@
+// From https://docs.microsoft.com/en-us/windows/desktop/dlls/creating-a-simple-dynamic-link-library
```
vitalych: Add a license header to this file.

tests/data/src/Myputs.c
```diff
@@ -0,0 +1,53 @@
...
+// Compile with: x86_64-w64-mingw32-gcc Myputs.c -o Myputs.dll -shared
```
vitalych: Could use all lower case for myputs.c and dll.

tests/data/src/Myputs.c
```diff
@@ -0,0 +1,53 @@
...
+
+    hConout = CreateFileW(L"CONOUT$",
+        GENERIC_WRITE,
+        FILE_SHARE_WRITE,
+        NULL,
+        OPEN_EXISTING,
+        FILE_ATTRIBUTE_NORMAL,
+        NULL);
+
+    if (INVALID_HANDLE_VALUE == hConout)
```
vitalych: Use {} for if.
Also run clang-format on this file.

vitalych:
We definitely need both the source and the binary. It would be preferable to build the DLL when running the tests, but it's probably easier to just have the binary for now, as long as it's clearly explained how to setup the compiler (all `apt-get install` commands, etc.).

vitalych:
Could you click on `Resolve` on every change you applied? I don't see some of them fixed.
Re. jenkins, yes, you need to change this file. I hope pytest returns a non-zero code in case of errors so that Jenkins can report the failure. You can deploy and test Jenkins locally on your own repos.

adrianherrera:
Oh is this "resolve conversation" a new feature? I swear I've never had to do it before....

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+  HANDLE hConout;
+  BOOL fRet;
+
+  // Get a handle to the console output device.
+
+  hConout = CreateFileW(L"CONOUT$", GENERIC_WRITE, FILE_SHARE_WRITE, NULL,
+                        OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+
+  if (INVALID_HANDLE_VALUE == hConout) {
+    return EOF;
```
vitalych: This doesn't look like it went through clang-format.
Only 2 spaces instead of 4.

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  // Write a null-terminated string to the console output device.
+
+  while (*lpszMsg != L'\0') {
+    fRet = WriteConsole(hConout, lpszMsg, 1, &cchWritten, NULL);
+    if ((FALSE == fRet) || (1 != cchWritten)) {
```
vitalych: Odd microsoft style, I'd write `fRet == FALSE`.

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+      return EOF;
+    }
+    lpszMsg++;
+  }
+  return 1;
```
vitalych: Return 0 might make more sense. Also ne dto add a new line after } before the return.

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include <windows.h>
+
+#define EOF (-1)
+
+#ifdef __cplusplus // If used by C++ code,
+extern "C" {       // we need to export the C interface
+#endif
+
+__declspec(dllexport) int __cdecl myPuts(LPWSTR lpszMsg) {
```
vitalych: myPuts case doesn't follow any convention.
For windows, it whould be `MyPuts` and the return type `INT`.

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+  hConout = CreateFileW(L"CONOUT$", GENERIC_WRITE, FILE_SHARE_WRITE, NULL,
+                        OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+
+  if (INVALID_HANDLE_VALUE == hConout) {
+    return EOF;
+  }
+
+  // Write a null-terminated string to the console output device.
+
+  while (*lpszMsg != L'\0') {
```
vitalych: I would get rid of all these ugly hungarian prefixes.

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+#define EOF (-1)
+
+#ifdef __cplusplus // If used by C++ code,
+extern "C" {       // we need to export the C interface
+#endif
+
+__declspec(dllexport) int __cdecl myPuts(LPWSTR lpszMsg) {
+  DWORD cchWritten;
+  HANDLE hConout;
+  BOOL fRet;
```
vitalych: `Written, Conout, Ret`

tests/data/src/myputs.c
```diff
@@ -0,0 +1,58 @@
...
+// method supported by Microsoft Visual Studio, but any
+// other export method supported by your development
+// environment may be substituted.
+
+#include <windows.h>
+
+#define EOF (-1)
+
+#ifdef __cplusplus // If used by C++ code,
+extern "C" {       // we need to export the C interface
```
vitalych: I think the comments are not useful here, it's obvious from the code.

tests/data/src/myputs.c
adrianherrera: Oh right, it turns out that the .clang-format file has to be in the same directory as the source code, **not** the same directory from where you run clang-format from. I've copied the version from build-scripts into tests/data/src

adrianherrera:
Done

Issue #228, opened 2019-01-01
s2e export_project does not export all files


vitalych:
E.g., in a driver project, sys files are missing.
It should zip all files in the project instead of using a whitelist.

adrianherrera:
Yeah my bad. A few things in export/import a broken, particularly with custom projects. Fixing it now. Expect a PR soon :)

PR #229, opened 2019-01-02
Fixes/import export

adrianherrera:
Fixes https://github.com/S2E/s2e-env/issues/228 and other issues associated with exporting custom projects.

There should really be some tests for this, but I'm not really sure how to write them.

### DIFF
--- a/s2e_env/commands/export_project.py
+++ b/s2e_env/commands/export_project.py
@@ -75,12 +75,12 @@ class Command(ProjectCommand):
         # Name the project archive if it doesn't already have a name
         output_path = options['output_path']
         if not output_path:
-            output_path = self.env_path('%s.tar.xz' % self._project_name)
+            output_path = self.env_path('%s.tar.xz' % self.project_name)
 
         with TemporaryDirectory() as temp_dir:
             # Store all of the exported files in a temporary directory so that
             # we can just execute tar on the entire directory
-            export_dir = os.path.join(temp_dir, self._project_name)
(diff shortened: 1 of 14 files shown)


s2e_env/commands/export_project.py
```diff
@@ -81,75 +81,62 @@ def handle(self, *args, **options):
...
-            # Copy project scripts and config files and rewrite the S2E
-            # environment path in these files
+            # Do **not** export these files
+            blacklist = CONSTANTS['import_export']['blacklist']
+            if not options['export_results']:
+                blacklist.extend(['s2e-last', 's2e-out-*'])
+
+            # Copy the project directory
+            logger.info('Copying project %s', self.project_name)
+            shutil.copytree(self.project_path(), export_dir,
```
vitalych: Is this going to resolve any symlinks in the folder?
The zip must not contain symlinks to files outside the project folder, these symlinks have to be turned into real files instead.

vitalych:
It think that this can be easily tested by creating a new project (e.g., a driver project), exporting it, then importing back. And then compare if the files in the imported project are all there.

s2e_env/commands/export_project.py
adrianherrera: Yep. `copytree` has a `symlinks=False` default argument, which means that "the contents and metadata of the linked files are copied to the new tree" [1]

[1] https://docs.python.org/2/library/shutil.html#shutil.copytree

adrianherrera:
Tested and it works!

PR #255, opened 2019-02-22
basic_block coverage: gets stuck in infinite loop

adrianherrera:
The binary search can get stuck in an infinite loop when `mid` is assigned to `lo` and the first two checks on the tb/bb coverage do not pass. In this case, `lo` will always be the same value as `mid`, which can never be greater than `hi` and therefore the `while` loop will never exit.

I have no idea if this is the correct solution, it's late on a Friday evening...


vitalych:
Maybe we could make it look like this: https://en.wikipedia.org/wiki/Binary_search_algorithm
I already see some differences in the loop header and we have some strange extra conditions.

vitalych:
Closing stale PR

PR #260, opened 2019-02-28
image_build: don't check kvm group if not building with KVM

adrianherrera:


vitalych:
I like my version better, no need to pass an extra var.
https://github.com/S2E/s2e-env/pull/253/commits/d215c0d1afbd3efc5f61fdab159bd1021b9b378a

adrianherrera:
Yeah I agree, yours is better. It's fine, it was 30 seconds of work 🙂

PR #290, opened 2019-04-23
Issue/283 protobuf

vitalych:


s2e_env/commands/forkprofile.py
```diff
@@ -25,11 +25,13 @@
 import logging
 import os
 
+from sortedcontainers import SortedDict
+
 from s2e_env.command import ProjectCommand, CommandError
 from s2e_env.execution_trace import parse as parse_execution_tree
+from s2e_env.execution_trace import TraceEntries_pb2
```
adrianherrera: Where is `TraceEntries_pb2`? I can't see it in this commit?

adrianherrera: Wait, nevermind, I think I am reading the commits in the wrong order.

PR #317, opened 2019-08-28
Fix an installation issue in with python3

benquike:
The repo can not be installed with pip in python3, because of incompatibilities in the setup.py of multiprocessing pip repo.

In python3, multiprocessing is preinstalled, thus does not need to be installed. This patch fixes the installation issue by removing multiprocessing from dependences in python3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ SOFTWARE.
 
 import os
 
+import sys
 import pip
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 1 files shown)


adrianherrera:
Hi Hui,

Thanks for this. Does the google repo tool work with python 3 yet? One of
the main hold ups of moving to python 3 was that the repo tool that we use
to manage all the different repos only worked on python 2.

~Adrian

benquike:
@adrianherrera I think so. 

https://gerrit.googlesource.com/git-repo/+/refs/heads/stable/repo#141

adrianherrera:
@benquike can you please sign the CLA and then I can merge.

vitalych:
CLA has not been signed. Closing.

PR #417, opened 2020-07-29
allow early termination when timeout is set

lzto:
Hi,
This patch allows early termination when execution time is less than timeout.
This patch is extremely useful and can save time when doing batch run.
Please consider pulling if you think this is useful.
Thanks.
- Tong

### DIFF
--- a/s2e_env/commands/run.py
+++ b/s2e_env/commands/run.py
@@ -106,10 +106,12 @@ def _sigterm_handler(signum=None, _=None):
 
 
 def _wait_for_termination(timeout):
+    cnt = timeout * 60
     while not terminating():
         if timeout:
-            time.sleep(timeout * 60)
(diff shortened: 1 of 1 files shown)


vitalych:
Thanks, it looks good. A couple fixes needed:
- Code style (see the build results)
- Prefix commit message with `commands/run:` so that it's easier to see what was changed by looking at the history.

lzto:
Thanks for the comments.
I did a couple of fixes as suggested. Please see if this version works.
Thanks,
- Tong

vitalych:
Cool, please also prefix the commit message with `commands/run:` so that it's easier to see later what was changed by looking at the history.

lzto:
hmm I see, prefix added.
Thanks,
- Tong

PR #449, opened 2021-12-07
s2e_env: fixed basic_block.py infinite loop

BMorgan1296:
Signed-off-by: Bradley Morgan <compsec@bradm.io>


BMorgan1296:
There were only two deprecation warnings from the output of `test.sh`, unrelated to my changes.

s2e_env/commands/code_coverage/basic_block.py
```diff
@@ -106,23 +106,23 @@ def _binary_search(tb_start_addr, bbs):
...
         mid = (lo + hi) // 2
 
         if bbs[mid - 1].end_addr < tb_start_addr <= bbs[mid].end_addr:
             return mid
         if tb_start_addr <= bbs[mid].end_addr:
-            hi = mid
+            hi = mid - 1
         else:
-            lo = mid
+            lo = mid + 1
```
vitalych, line 125: Thanks, it looks better than this previously abandonned PR.
https://github.com/S2E/s2e-env/pull/255/commits/444a884d2af57552bbcddcdaf7c0fac9b557a59d

Could you make all the loop conditions exactly like on wikipedia? Something like this:

```
if A[m] < T then
    L := m + 1
else if A[m] > T then
    R := m − 1
else:
    return m
``` 

Currently we have `bbs[mid - 1].end_addr < tb_start_addr <= bbs[mid].end_addr:` and `tb_start_addr <= bbs[mid].end_addr` and I can't easily reason about its correctness.

A unit test would also be useful.

BMorgan1296: How would you like the unit test to be structured? I've been using Radare2 as my disassembler. Is it suitable to force its use in such a test?

vitalych: No, using the real radare2 would be an integration test. For the unit test, you'd just have to simulate the input to the function that causes the bug. You could dump it by printing the arguments that caused the original function to hang.

s2e_env/commands/code_coverage/basic_block.py
```diff
@@ -305,7 +305,8 @@ def _get_cached_disassembly_info(self, module):
         # modification time compared to the .disas file
 
         disas_mtime = os.path.getmtime(disas_path)
-        target_mtime = os.path.getmtime(self._project_desc['target_path'])
+        target_mtime = os.path.getmtime(self._project_desc['project_dir'])
+        #print(dir(self._project_dir))
```
vitalych, line 309: Don't leave commented out code

tests/commands/code_coverage/test_bb_code_coverage_elf.py
```diff
@@ -0,0 +1,46 @@
...
+import itertools
+from unittest import TestCase
+
+from s2e_env.commands.code_coverage import basic_block
+
+
+class ELFTestCase(TestCase):
+    def test_bb_cov(self):
+        """Test basic block code coverage of an x86 ELF executable, mmount"""
+        file1 = open('tests/commands/code_coverage/mmount_bbs.obj', 'rb')
```
vitalych: Would it be possible to avoid pickled files? Besides security issues, they are hard to edit. Use Python code instead.

tests/commands/code_coverage/test_bb_code_coverage_elf.py
```diff
@@ -26,16 +26,18 @@
...
 class ELFTestCase(TestCase):
     def test_bb_cov(self):
         """Test basic block code coverage of an x86 ELF executable, mmount"""
-        file1 = open(os.getcwd()+'/tests/commands/code_coverage/mmount_bbs.obj', 'rb') 
-        file2 = open(os.getcwd()+'/tests/commands/code_coverage/mmount_tb_coverage.obj', 'rb')
-        bbs = pickle.load(file1)
-        tb_coverage = pickle.load(file2)        
-        file1.close()
-        file2.close()
+        bbs = tb_coverage = None
```
vitalych, line 35: Squash this commit. Pull requests should not have fixes to their own code.

vitalych (changes requested):
see comments

tests/commands/code_coverage/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+
+
+import os
+
+CURR_DIR = os.path.join(os.getcwd(), os.path.dirname(__file__))
```
vitalych, line 26: I wouldn't use getcwd(), it forces the tests to be run from a specific directory.

tests/commands/code_coverage/test_bb_code_coverage_elf.py
BMorgan1296: Done.

vitalych:
Force-pushing is the proper method to handle this. It's your private branch, so force-push is fine. It's important to do this because it makes it easier for reviewers. A reviewer might spend time commenting on some code issue only to find in a later commit that this issue was resolved. 

Note that you shouldn't necessarily force-push everything into one commit either. They still must be logically split, again, to make it easier to review. In this particular case, if you find yourself calling the commit `fixed basic_block.py infinite loop, added unit test, fixed a dict field error`, you would probably need one commit for each fix:
1. infinite loop
2. dict error
3. added unit tests

tests/commands/code_coverage/mmount_tb_coverage.json
```diff
@@ -0,0 +1 @@
...392, 31442, 55], [31417, 31442, 30], [31447, 31457, 16], [31463, 31465, 8], [31471, 31483, 17], [31488, 31502, 19], [31507, 31521, 19], [31526, 31528, 8], [31534, 31534, 5], [31539, 31545, 8], [31547, 31559, 17], [31564, 31590, 31], [31595, 31604, 14], [31609, 31668, 62], [31668, 31668, 3], [31671, 31678, 12], [31683, 31689, 8], [31698, 31712, 16], [31751, 31758, 13], [31764, 31767, 8], [31772, 31778, 8], [31912, 31920, 10], [31932, 31942, 11], [31952, 31964, 17], [31969, 31981, 17], [31986, 31986, 5], [32000, 32022, 27], [32027, 32027, 5], [101744, 101759, 20], [101764, 101767, 5], [101769, 101773, 5], [103200, 103213, 18], [103218, 103233, 16], [111824, 111880, 58], [111882, 111893, 13], [111895, 111920, 30], [111925, 111930, 10], [111935, 111939, 6], [130368, 130390, 24], [130416, 130416, 5], [130421, 130431, 15], [130512, 130522, 15], [130527, 130535, 13], [130540, 130548, 13], [130553, 130559, 11], [130564, 130572, 10], [130586, 130588, 4], [130594, 130600, 11], [131088, 131108, 25], [131113, 131125, 17], [131130, 131142, 17], [131147, 131163, 21], [144448, 144492, 49], [144497, 144501, 6], [144503, 144521, 22], [144525, 144532, 9], [144534, 144548, 15], [144596, 144608, 13]]}
```
vitalych, line 1: This looks good

tests/commands/code_coverage/test_bb_code_coverage_elf.py
```diff
@@ -0,0 +1,46 @@
...
+from s2e_env.commands.code_coverage import basic_block
+
+from . import CURR_DIR
+
+class ELFTestCase(TestCase):
+    def test_bb_cov(self):
+        """Test basic block code coverage of an x86 ELF executable, mmount"""
+        bbs = tb_coverage = None
+
+        with open(CURR_DIR+'/mmount_bbs.json', 'r') as f1:
```
vitalych, line 35: You might need to add `encoding='utf-8'` to open. I was getting pylint warnings.

vitalych, line 35: `os.path.join()` would be better

vitalych (approved):
one last small remark

PR #456, opened 2022-01-20
Remove pygit2 dependency

### LINKED ISSUE #455: Remove pygit2 dependency
We are trying to get s2e-env installed on Ubuntu 20.04, the latest LTS release. We've run into a version conflict on pygit2, which is pinned to 1.2.1, and requires libgit2 v1.0.X. Ubuntu's libgit2-dev package is v0.28, which is incompatible and pygit2 fails to build.

Doing a quick search, it looks like pygit2 is only used in s2e-env to [get the user name when creating a plugin](https://github.com/S2E/s2e-env/blob/master/s2e_env/commands/new_plugin.py#L65). I purpose that this code changes slightly:

- Add an optional `--author-name` parameter
- If `--author-name` isn't provided, use the current user name, via something like [`getpass.getuser()`](https://docs.python.org/3/library/getpass.html#getpass.getuser)
- Remove the pygit2 dependency

If this issue is approved, I should be able to handle the PR since I think it's straightforward.

ameily:
This PR removes the pygit2 dependency, which was previously used to get the plugin author's full name in the `new_plugin` command. Now, the command gets the username from either the optional `--author-name` command line argument or, as a fallback, through `getpass.getuser()`.

fixes #455

### DIFF
--- a/s2e_env/commands/new_plugin.py
+++ b/s2e_env/commands/new_plugin.py
@@ -64,7 +64,7 @@ def _inject_plugin_path(makefile, plugin_path):
 def _get_user_name():
     try:
         username = subprocess.check_output(['git', 'config', 'user.name']).decode()
-    except:
+    except subprocess.CalledProcessError:
         username = ''
 
     return username.strip()
@@ -90,7 +90,7 @@ class Command(EnvCommand):
         parser.add_argument('--force',
                             help='Overwrites existing plugin',
                             action='store_true')
-        
+
         parser.add_argument('--author-name',
                             help='The plugin author name (uses git config "user.name" if not provided).')
 
--- a/s2e_env/commands/new_plugin.py
+++ b/s2e_env/commands/new_plugin.py
@@ -23,8 +23,7 @@ SOFTWARE.
 import datetime
 import logging
 import os
-
-import pygit2
+import subprocess
 
 from s2e_env.command import EnvCommand, CommandError
 from s2e_env.utils.templates import render_template, DEFAULT_TEMPLATES_DIR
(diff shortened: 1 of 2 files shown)


s2e_env/commands/new_plugin.py
```diff
@@ -91,13 +82,21 @@ def add_arguments(self, parser):
...
+                            help='The plugin author name (uses git config "user.name" if not provided).')
+
     # pylint: disable=too-many-locals
     def handle(self, *args, **options):
         s2e_src_dir = self.env_path('source', 's2e', 'libs2eplugins', 'src')
         s2e_plugins_dir = self.env_path(s2e_src_dir, 's2e', 'Plugins')
 
         plugin_name = os.path.basename(options['plugin_name'][0])
         plugin_rel_dir = os.path.dirname(options['plugin_name'][0])
+        author = options['author_name'] or getpass.getuser()
```
vitalych, line 95: Why not calling `git config user.name` directly instead, then save the output?
`getpass.getuser()` gets the user account name, which is not the full name.

ameily:
@vitalych I believe I addressed your feedback.

vitalych:
Looks good. Couple more remarks:
- Squash the pylint fixes with the previous commits. In general, a pull request should not have commits that fix earlier commits of that PR. This helps with review, i.e., the reviewer might spend some time commenting on a commit only to find out in later commits that the comment was addressed.
- Add prefixes to the commit like in my comment above. This helps quickly figure out which module is affected by a commit when looking through the git history.

ameily:
I've done as you requested.

vitalych:
Thanks a lot!

PR #483, opened 2022-10-30
Update README.md

zijiancogito:


README.md
```diff
@@ -10,7 +10,7 @@ might be the analysis of the ``file`` program from
 
 # Prerequisites
 
-We assume that you are working on an Ubuntu 18.04 LTS 64-bit OS.
+We assume that you are working on an Ubuntu 20.04 LTS 64-bit OS.
```
vitalych, line 13: I would rephrase it like this: `We assume that you are working on the latest LTS release of Ubuntu on a 64-bit OS.`.
Otherwise, we'll need to keep updating this forever.

vitalych (changes requested):
Also change the commit message to: `readme: updated OS requirements`

0xk4n3ki:
lol, now its 22.04 :)

PR #510, opened 2024-12-05
S2E kprobe support + alternative image generation

m4mbr3:
This commit adds to s2e new_project the option --debootstrap flag.

The main job of this flag is to default to the generic debootstrap image description in images.json database independently from the name of the image passed through -i option. For debootstrap-based images, the folder name is set by the script as ${distro_name}-${kernel_version} to being able to express all possible combinations distro/kernel which was not a supported possibility by s2e new_project. Another change integrated by --debootstrap is to generate a different boostrap.sh script to insert the kprobe based kernel module to add s2e support to the untouched kernel. The s2e-kprobe kernel module sits in
/root/s2e-kprobe/s2e-`uname -r`/s2e-kprobe.ko. If changes to the kernel module are necessary (e.g., to add some s2e callbacks at the beginning of the analysis to share some kernel info), it is possible to override the current installed module through s2ecmd get without rebuilding the image.

### DIFF
--- a/s2e_env/templates/plugin_creation/plugin.cpp.template
+++ b/s2e_env/templates/plugin_creation/plugin.cpp.template
@@ -43,7 +43,7 @@ class {{ plugin.name }}State: public PluginState {
     // Declare any methods and fields you need here
 
 public:
-    static PluginState *factory(Plugin *p, S2EExecutionState *s) {
+    static PluginState *factory(Plugin *p) {
         return new {{ plugin.name }}State();
     }
 


s2e_env/templates/bootstrap.linux.sh
```diff
@@ -30,7 +30,11 @@ function execute_target {
 # Nothing more to initialize on Linux
 function target_init {
     # Start the LinuxMonitor kernel module
+    {% if debootstrap == true %}
```
vitalych, line 33: I think bootstrap.sh could auto-detect the type of Linux distro it's running on. If the kprobe module exists, use it instead of s2e. That would avoid having to pass the debootstrap around.

